Issue #297, opened 2021-11-04
Unable to run Linux barrier client during startup using systemd


ghost:
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/297
> 
> Issue created on: 2019-05-01 by @vnktsh
> Issue last updated on: 2021-03-16

### Operating Systems ###

Server: Windows 10 Version 1709 (OS Build 16299.1087)
Client: Ubuntu 18.04 (GNOME 3 desktop)

### Barrier Version ###
Server: 2.2.0
Client: 2.2.0

### Steps to reproduce bug ###

1. Goal is to make Linux barrier client to startup during boot
2. Setup systemd srevice script [attached service script below]
3. `sudo systemctl start barrier.service`
3. Barrier manages to startup but throws error: "**WARNING: secondary screen unavailable: unable to open screen**"
4. Fails to connect to server

### Other info ###

* When did the problem start to occur?
When trying to setup systemd service.
* Is there a way to work around it?
Not, if you want barrier to startup before logging in the user.
* Does this bug prevent you from using Barrier entirely?
No

**Put anything else you can think of here.**

My service script under `/etc/systemd/system/barrier.service`
```
[Unit]
Description=Start Barrier client during boot

[Service]
Type=simple
Restart=always
RestartSec=2
ExecStart=/usr/bin/barrierc -f --debug DEBUG2 --log /tmp/barrier-service.log --name ubuntu-Desktop [<server-info>]:24800

[Install]
WantedBy=multi-user.target
```

**Logs:**

```
tail -f /tmp/barrier-service.log 

[2019-04-30T22:26:50] DEBUG: XOpenDisplay(":0.0")
[2019-04-30T22:26:50] WARNING: secondary screen unavailable: unable to open screen
[2019-04-30T22:26:50] DEBUG: retry in 60 seconds
[2019-04-30T22:27:50] DEBUG: XOpenDisplay(":0.0")
[2019-04-30T22:27:50] WARNING: secondary screen unavailable: unable to open screen
[2019-04-30T22:27:50] DEBUG: retry in 60 seconds

```

---

> Commented on: 2019-05-01 by @noisyshape

If I had to guess I would say that it's running too early and Barrier doesn't correct for it. Using your display manager to start and stop Barrier might work. Or instead of running Barrier as root you could log in automatically. Pick your poison.

See also #179 and #185

---

> Commented on: 2019-05-03 by @vnktsh

[...]

PR #694, opened 2021-11-04
Add systemd services (resolves #36, resolves #508)

### LINKED ISSUE #36: Include a systemd user service for linux builds
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/36
> 
> Issue created on: 2018-05-09 by @AdrianKoshka
> Issue last updated on: 2019-11-29

I've written some systemd user services for my own personal use, and know some people might appreciate having a unit provided by upstream. This is mostly just a note for myself, I'll create a PR later, if you'd be open to the idea.

---

> Commented on: 2018-05-10 by @walker0643

Sure, Adrian. Would it be as simple as making a dist/archlinux dir to hold the template and then having cmake fill it in at config-time?

---

> Commented on: 2018-09-08 by @walker0643

Closing since we dropped this back in May. Go ahead and submit the PR if you still want to.

---

> Commented on: 2019-11-29 by @shymega

Reopening for the benefit of #508 and other users.

### LINKED ISSUE #508: Automatically start Barrier on boot in Raspbian Buster
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/508
> 
> Issue created on: 2019-11-22 by @ElnuDev
> Issue last updated on: 2021-05-23

### Operating Systems ###

Server: Windows 10

Client: Raspbian Buster

### Barrier Version ###

Server: 2.3.2-snapshot-210c2b70

Client: 2.3.2-snapshot-36ed253e

### Problem ###

I would like to make Barrier automatically start on boot on my Raspberry Pi 4 running Raspbian Buster. Currently, I have to connect a second mouse/keyboard pair to my Raspberry Pi in order to start up Barrier, which kind of defeats the purpose. 🙃 Is this possible? I tried adding `barrier &` to `etc/rc.local` and rebooting the Pi, but that didn't work.

---

> Commented on: 2019-11-29 by @shymega

I don't believe that Debian Buster uses `sysvinit`, and since Debian Jessie, it now uses systemd. 

#36 would have fulfilled this requirement, and I will make a PR soon to provide these units for a systemd user service, and _possibly_ a systemd system service, but I'm not really convinced we need a systemd system service.

Could you confirm that Raspbian Buster uses systemd or sysvinit? You can check for systemd by running `systemctl`, and seeing if returns an error.

Thanks for the report!

---

> Commented on: 2019-12-29 by @candeira

@Shymega, wouldn't a systemd system service allow users to use barrier for accessing client computers' display manager/login?

---

> Commented on: 2019-12-29 by @shymega

@candeira Not sure. It could potentially work, but it wouldn't be per-user, and only for the root user, I'd imagine - so you'd need two Barrier configs.. on the same system. At least, that's how I understand it. I'd personally stick with a user service, but I can see the appeal of Barrier being available on the DM.

---

> Commented on: 2020-01-07 by @mirh

My personal situation is a PC where I don't have physical keyboards attached, and in come cases I find myself having to switch to TTY. 
If the thing is running as a normal user.. Of course I cannot login into the console to begin with. 

---

> Commented on: 2020-01-17 by @dayne

My solution for this is a set of scripts used on my Raspberry Pi 4 and couple another Linux boxes that I revision control in a gist over here:

https://gist.github.com/dayne/e3a7f31f0624bf299faf9fadfe510322

I see I have a note about fixing up my bash script stuff to make it work correctly in a mac... I'll loop back to that in the future but what I've linked above works for me on my RPi.

---

> Commented on: 2020-02-03 by @detly

Before I go off and try to configure this from scratch, what *is* the expected way to initiate a Barrier connection with a headless client?

---

> Commented on: 2020-05-12 by @rebroad

Adding:-

`barrier &`

to the end of /home/pi/.profile was enough to get it running automatically upon boot up into desktop mode for me

---

> Commented on: 2020-05-12 by @detly

@rebroad In this context, "at boot" means before logging in.

---

> Commented on: 2020-05-27 by @detly

@lornix This will start Barrier on login, not on boot. This issue is about starting Barrier before any user is logged in, so that on multi-user systems someone can enter credentials at the login screen. 

---

> Commented on: 2020-05-28 by @simons-public

@detly I put in a PR #694 that has systemd services if you want to give that a try.

---

> Commented on: 2020-05-28 by @detly

@simons-public Amazing! I left a review comment, I hope you don't mind me jumping in.

---

> Commented on: 2020-06-10 by @Zazaaaa

> Adding:-
> 
> `barrier &`
> 
> to the end of /home/pi/.profile was enough to get it running automatically upon boot up into desktop mode for me

I am sorry, but could someone please explain to me how I am supposed to do this? I am completely starting from scratch with my Raspberry Pi and I want to connect my keyboard and mouse first. I don't have any other users but the main one so I don't need to log on. I just need to run Barrier on startup so I don't have to plug in my keyboard first. Can somebody please help me? Thanks in forward.

---

> Commented on: 2020-06-12 by @lachlanshanks

> > Adding:-
> > `barrier &`
> > to the end of /home/pi/.profile was enough to get it running automatically upon boot up into desktop mode for me
> 
> I am sorry, but could someone please explain to me how I am supposed to do this? I am completely starting from scratch with my Raspberry Pi and I want to connect my keyboard and mouse first. I don't have any other users but the main one so I don't need to log on. I just need to run Barrier on startup so I don't have to plug in my keyboard first. Can somebody please help me? Thanks in forward.

If you visit the /home/pi folder in the file explorer and in the drop down menu beside the address bar select to show hidden files, you'll see the ".profile" file. You just need to add "barrier &" to the end of that.

EDIT: This is working for me now.

---

> Commented on: 2020-06-25 by @julio8a

Thank you! adding `barrier &` to the end of the `/home/pi/.profile` file worked for me. 

---

> Commented on: 2020-06-25 by @Sk8wine

> Adding:-
> 
> `barrier &`
> 
> to the end of /home/pi/.profile was enough to get it running automatically upon boot up into desktop mode for me

Thank you! Was forever changing dongle to boot Barrier - worked perfectly

---

> Commented on: 2020-06-28 by @LRC32901

I've added "barrier & to the end of my " .profile" file
When screen boots I get a screen looking for Server IP: .
Is there a way to pass it the ip or did I miss something on setuo.
Thanks

---

> Commented on: 2020-07-19 by @skorokithakis

@LRC32901 You may be able to change the line to `barrierc <server ip> &` but I haven't tested that.

---

> Commented on: 2020-10-08 by @blackguard

This is working for me:
1) create a script (below) to start barrierc and connect to the local barrier server.
2) call that script from the end of ~/.profile

The script (change barrier_host to the correct ip address for your barrier server, i.e., the one with the keyboard and mouse):

#!/bin/bash
export barrier_host=192.168.1.1  # CHANGE ME
export barrier_port=24800
export barrier_logfile="${HOME}/barrier.log"
if ! pgrep -af /usr/bin/barrierc; then
    nohup /usr/bin/barrierc --no-tray --debug INFO --name raspberrypi --enable-crypto "[${barrier_host}]:${barrier_port}" <&- > "${barrier_logfile}" 2>&1 &
fi

---

> Commented on: 2020-12-11 by @DanielGlaas

Hi guys, here is my solution to have the client automatically running on my raspberry pi:

edit file 
`sudo nano /etc/xdg/lxsession/LXDE-pi/autostart`
and add this line
`@barrierc -f --name raspberrypi --enable-crypto 192.168.0.2
`

The standard user "pi" is automatically loged in, this means that on the monitor already the desktop is visible. As soon as I start barrier server on my Laptop with Ubuntu 20.04, the connection establishes within seconds.

Maybe this helps someone!

---

> Commented on: 2021-01-22 by @Kingo71

> Hi guys, here is my solution to have the client automatically running on my raspberry pi:
> 
> edit file
> `sudo nano /etc/xdg/lxsession/LXDE-pi/autostart`
> and add this line
> `@barrierc -f --name raspberrypi --enable-crypto 192.168.0.2 `
> 
> The standard user "pi" is automatically loged in, this means that on the monitor already the desktop is visible. As soon as I start barrier server on my Laptop with Ubuntu 20.04, the connection establishes within seconds.
> 
> Maybe this helps someone!

This worked like a charm for me, thanks!

---

> Commented on: 2021-03-29 by @1oh1

> Hi guys, here is my solution to have the client automatically running on my raspberry pi:
> 
> edit file
> `sudo nano /etc/xdg/lxsession/LXDE-pi/autostart`
> and add this line
> `@barrierc -f --name raspberrypi --enable-crypto 192.168.0.2 `
> 
> The standard user "pi" is automatically loged in, this means that on the monitor already the desktop is visible. As soon as I start barrier server on my Laptop with Ubuntu 20.04, the connection establishes within seconds.
> 
> Maybe this helps someone!

If you use Dynamic DNS providers like No-IP, DuckDNS or similar, you can replace the IP address in the command with the your DDNS hostname (eg. `my-awesome-pc.duckdns.org`) as long as the hostname points to your computer's local IP and not public IP.

---

> Commented on: 2021-05-23 by @ericlowry

> Hi guys, here is my solution to have the client automatically running on my raspberry pi:
> 
> edit file
> `sudo nano /etc/xdg/lxsession/LXDE-pi/autostart`
> and add this line
> `@barrierc -f --name raspberrypi --enable-crypto 192.168.0.2 `
> 
> The standard user "pi" is automatically loged in, this means that on the monitor already the desktop is visible. As soon as I start barrier server on my Laptop with Ubuntu 20.04, the connection establishes within seconds.
> 
> Maybe this helps someone!

@DanielGlaas, I cam looking for exactly this - worked like a charm!  Thanks for saving hours of digging.

ghost:
> This PR has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/pull/694
> 
> PR created on: 2020-05-26 by @simons-public
> PR last updated on: 2021-07-25

This pull request adds 3 systemd services:

`barriers.service` - Starts the Barrier server with defaults
`barriers@.service` - Can be used to start a Barrier server instance on another port
`barrierc@.service` - Used to start Barrier client instance

The server service will create a `Barrier.pem` self-signed certificate on first run (all other certificate generation in Barrier happens in the GUI). The server instance (with the @) allows starting with a specific listening port with `systemctl start barriers@:<port>' or a specific listening IP/port with `systemctl start barriers@<ip>:<port>`.

The client service will create a 'TrustedServers.txt' with the specified host on first run. It is started as an instance with `systemctl start barrierc@<hostname>` or enabled with `systemctl enable barrierc@<hostname>`. The port can also be specified (i.e. `barrierc@<hostname>:<port>`).

Since these are for using barrier system-wide via systemd, the configuration file specified is `/etc/barrier.conf` and the data is stored in `/var/db/barrier*` instead of `/root/.local/share/barrier`.

They are installed with CMake in `/usr/lib/systemd/system` so they can be overridden and customized by creating a copy placed in `/etc/systemd/system/`.

---

> Commented on: 2020-05-26 by @p12tic

What distribution was this default picked from? Wouldn't it better to use `/var/lib` which is standard?

---

> Commented on: 2020-05-26 by @p12tic

I think it would be better to use a single directory for all barrier data, e.g. `/var/lib/barrier/hostname-<hostname>`.

---

> Commented on: 2020-05-26 by @p12tic

A hostname of the server is encoded here... What happens if it is offline?

---

> Commented on: 2020-05-26 by @p12tic

This will delete all other keys in the file.

---

> Commented on: 2020-05-26 by @simons-public

Sorry that was an oversight, I'll replace that with the correct instance variable

---

> Commented on: 2020-05-26 by @simons-public

I'll change it to `/var/lib`

---

> Commented on: 2020-05-26 by @simons-public

The `[ -f "${FP_DIR}/TrustedServers.txt" ] || command` should only run if `[ -f` fails which would mean there isn't an existing file.

---

> Commented on: 2020-05-26 by @simons-public

Edit: I think I misunderstood what you were saying, I'll make this change as well

---

> Commented on: 2020-05-26 by @simons-public

@p12tic Thanks for the review! I pushed the requested change to `/var/lib`, put all barrier data into a single directory, and got rid of the hardcoded hostname (oops). The other item I added an explanation for.

---

> Commented on: 2020-05-27 by @p12tic

Typo.

---

> Commented on: 2020-05-27 by @p12tic

I think this should be `/var/lib/barrier/barrier<instance>/barrier/SSL`

---

> Commented on: 2020-05-27 by @p12tic

Just two tiny nitpicks left, otherwise good. Thank you!

---

> Commented on: 2020-05-27 by @p12tic

Sounds reasonable.

---

> Commented on: 2020-05-28 by @simons-public

@p12tic Got them fixed, thanks again!

---

> Commented on: 2020-05-28 by @detly

Note that if Barrier is installed via Snap, the path to the executable is `/snap/bin/barrier.barrierc` etc. Are there CMake tokens for this you could use?

---

> Commented on: 2020-05-29 by @detly

Note that Systemd `Exec*` commands need to be an absolute path on the current LTS release of Ubuntu (18.04.3) which runs Systemd 237. From that version of the man page ("COMMAND LINES" section):

> The command to execute must be an absolute path name. It may contain spaces, but control characters are not allowed.

The current version of Systemd [seems to relax this](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines):

> If the command is not a full (absolute) path, it will be resolved to a full path using a fixed search path determinted at compilation time. Searched directories include /usr/local/bin/, /usr/bin/, /bin/ on systems using split /usr/bin/ and /bin/ directories, and their sbin/ counterparts on systems using split bin/ and sbin/. It is thus safe to use just the executable name in case of executables located in any of the "standard" directories, and an absolute path must be used in other cases. Using an absolute path is recommended to avoid ambiguity. Hint: this search path may be queried using systemd-path search-binaries-default.

---

> Commented on: 2020-05-29 by @detly

(Also, if you are concerned with targeting Ubuntu 18.04.3, which you understandably might not be, the `ServiceType=exec` will be invalid; you will need to use `ServiceType=simple`.)

---

> Commented on: 2020-05-29 by @detly

I've been playing around with this, running it on an Ubuntu 18.04.3 with Barrier installed via Snap (which might not be a case you want to support). I keep running into #297 . See also the linked issues there, which also mention problems with running it as root.

(Digging into it it seems like this has nothing to do with Snap, since the same problem is reported for a variety of installation sources and also for Synergy/Symless.)

---

> Commented on: 2020-05-29 by @detly

Another note re. Snap (caveat: I know almost nothing about Snap). If `barrierc` is to run as root, the Snap apparmor policy `/var/lib/snapd/apparmor/profiles/snap.barrier.barrierc` will probably need to change, in particular access to `/root` and wherever the `Xauthority` file is (maybe?).

IIRC Systemd *also* has the potential to run a service as a dynamic, throwaway user. That might be preferable.

---

> Commented on: 2020-05-29 by @detly

The service still creates `/var/lib/barrier/barrier@<server>/barrier/SSL/Fingerprints/` directories fine, but I'm seeing apparmor denials for `/root/snap/barrier/83/.local/share/barrier/SSL/Fingerprints/TrustedServers.txt`. The service file sets `XDG_DATA_HOME=/var/lib/barrier/barrier@%i`, but is Barrier using `$HOME` instead?

Update: it was [this commit](https://github.com/debauchee/barrier/commit/642eb334462f9d0433f9f7028c8b054c22dffe1a) that made the change, and the same code uses `.local/share/barrier` instead of `.barrier`, so I'm a bit baffled by this.

`journalctl` shows `NOTE: trustedServersFilename: /root/snap/barrier/83/.local/share/barrier/SSL/Fingerprints/TrustedServers.txt` which comes from [here](https://github.com/debauchee/barrier/blob/master/src/lib/net/SecureSocket.cpp#L704) in the code. No idea why it's not actually picking up `XDG_DATA_HOME`.

---

> Commented on: 2020-05-29 by @simons-public

I'm not seeing any thing in the CMakeLists.txt, but I'm doing some reading on Snap to see if there's an environment variable or something that can be used to change the path with CMake.

---

> Commented on: 2020-05-29 by @simons-public

Good point, I'll change it to absolute paths

---

> Commented on: 2020-05-29 by @detly

Okay, see [Snap forum post](https://forum.snapcraft.io/t/is-xdg-data-home-modified-by-snaps-adapters/17826).

Unrelated: I have also found that I need:

```text
Environment=XAUTHORITY=/var/run/lightdm/root/:0
```

...and the appropriate lines in `/var/lib/snapd/apparmor/profiles/snap.barrier.barrierc` to allow access to that. It seems very fragile eg. what if the user is using GDM or KDM or whatever? But I tried years ago to work out a generic ["where is the Xauthority file" mechanism](https://lists.ubuntu.com/archives/ubuntu-devel/2013-November/037913.html) with no results :grimacing:.

---

> Commented on: 2020-05-29 by @simons-public

@detly Thanks for all the feedback. 

Using `simple` instead of `exec` works just fine, I don't mind changing that at all.

As far as #297 goes I think it might have something to do with Xauthority, I'll spin up an Ubuntu VM this weekend to see if I can replicate and figure it out.

The path search happens in [DataDirectories.cpp](https://github.com/debauchee/barrier/blob/dbd10820c3090ef72cb088023a8bf2249450c8b5/src/lib/common/unix/DataDirectories.cpp#L63-L70) and falls back to using $HOME, so it looks like for some reason it's not reading the $XDG_DATA_HOME environment variable. Maybe Snap is unsetting $XDG_ variables, I'll have to keep reading up on it.

I'm not particularly fond of the idea of running barrier as root anyways, so I'll do some playing around with using those systemd dynamic users. I think the only difficult part of implementing that might be granting the user permissions to the X11 display (possibly using `xhost` or something). 

---

> Commented on: 2020-05-29 by @detly

@simons-public We might have crossed signals there, see my last post. As an addendum, if you find [this hack](https://lists.ubuntu.com/archives/ubuntu-devel/2013-February/036486.html) don't get your hopes up, that ACPI stuff has been deprecated for years.

---

> Commented on: 2020-05-29 by @detly

Re. running as root... I wonder if it might be possible to run as whatever user the display manager runs as? And then there might be a way to get `XAUTHORITY` more easily?

If I get more time to work on this, I'll be sure to sync up with you so we don't duplicate effort.

---

> Commented on: 2020-05-29 by @simons-public

@detly Thanks for the tip on not going down the acpi rabbit hole.

It should be possible to determine which user is running X11 by looking at the owner of the `/tmp/.X11-unix/X0` socket, I'm just not sure how portable that path is between distros. I haven't read up yet if there is any way to dynamically set the User= for a systemd service.

---

> Commented on: 2020-05-29 by @p12tic

> I'm just not sure how portable that path is between distros

systemd integration is distribution-specific anyway. I think a good enough solution would be look into what Debian, Ubuntu, Centos and Fedora uses and handle those cases. Whoever uses other distributions can implement support for their specific distribution and submit a bugfix upstream.

---

> Commented on: 2020-05-30 by @shymega

> > I'm just not sure how portable that path is between distros
> 
> systemd integration is distribution-specific anyway. I think a good enough solution would be look into what Debian, Ubuntu, Centos and Fedora uses and handle those cases. Whoever uses other distributions can implement support for their specific distribution and submit a bugfix upstream.

Agreed. Also note that distributions can override parts of the supplied unit if need be.

---

> Commented on: 2020-06-01 by @jmartens

Very nice to see systemd implementation. 

Not sure if I should raise it here, as a bug or that I am doing something wrong. I tried it on my local build, but after a `make install` it does not seem to start:

```
  build git:(systemd-services) ✗ sudo systemctl start barriers         
Job for barriers.service failed because the control process exited with error code.
See "systemctl status barriers.service" and "journalctl -xe" for details.
➜  build git:(systemd-services) ✗ sudo systemctl status barriers.service
● barriers.service - Barrier Server listening on 24800 (Open-source KVM software)
     Loaded: loaded (/usr/local/lib/systemd/system/barriers.service; disabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Mon 2020-06-01 10:36:12 CEST; 4s ago
       Docs: man:barriers(1)
             man:barrierc(1)
             https://github.com/debauchee/barrier/wiki
    Process: 55798 ExecStartPre=/usr/bin/mkdir -p ${CERT_DIR} (code=exited, status=203/EXEC)

jun 01 10:36:11 laptop-mapa systemd[1]: barriers.service: Control process exited, code=exited, status=203/EXEC
jun 01 10:36:11 laptop-mapa systemd[1]: barriers.service: Failed with result 'exit-code'.
jun 01 10:36:11 laptop-mapa systemd[1]: Failed to start Barrier Server listening on 24800 (Open-source KVM software).
jun 01 10:36:12 laptop-mapa systemd[1]: barriers.service: Scheduled restart job, restart counter is at 5.
jun 01 10:36:12 laptop-mapa systemd[1]: Stopped Barrier Server listening on 24800 (Open-source KVM software).
jun 01 10:36:12 laptop-mapa systemd[1]: barriers.service: Start request repeated too quickly.
jun 01 10:36:12 laptop-mapa systemd[1]: barriers.service: Failed with result 'exit-code'.
jun 01 10:36:12 laptop-mapa systemd[1]: Failed to start Barrier Server listening on 24800 (Open-source KVM software).
➜  build git:(systemd-services) ✗ 
```

---

> Commented on: 2020-06-01 by @jmartens

>     Process: 55798 ExecStartPre=/usr/bin/mkdir -p ${CERT_DIR} (code=exited, status=203/EXEC)

If I manually run all ExecStartPre as root, disable the ExecStartPre statements in the script and run `systemctl daemon-reload` I am able to get past them only to discover the install path is not reflected properly in the services file.

Since I use make install with an install target it should probably reflect that install target as well.

After manually changing the path to the barrier executable I run into the following issue a missing /etc/barrier.conf file. Should we perhaps provide a default one?

---

> Commented on: 2020-06-01 by @maxiberta

Snaps have native support for daemon services. Would that help?
https://snapcraft.io/docs/services-and-daemons

---

> Commented on: 2020-06-05 by @simons-public

@jmartens I think the issue with the  `mkdir` command could be [permissions](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStart=) which might be fixed by prepending a `+` symbol on the command line. Are you running `systemctl` as a user or with elevated privileges (i.e `sudo`)? I'm haven't gotten around to snap paths yet, but I'll test and try to push a fix for the `ExecStartPre` when I get home. 

I don't know if it would make sense to have a default `barrier.conf` file since the configuration items are so specific to individual setups, but there are example ones in the [doc](https://github.com/debauchee/barrier/tree/master/doc) dir of the repo.

---

> Commented on: 2020-06-06 by @simons-public

@detly Would you mind testing if the snap paths work on your setup with 368f0d7? I haven't had time to get a VM setup but I added some checks for snapcraft environment variables in the CMakeLists.txt.

@jmartens Can you let me know if you're having the same issues after 95ffe58?

---

> Commented on: 2020-06-06 by @detly

@simons-public can do! I'll get back to you in a few hours. 

---

> Commented on: 2020-06-06 by @simons-public

@jmartens I didn't know about those items, thanks for sharing. I tested it out and it works well so I replaced all the mkdir commands with sandboxing directory directives in 35ac344 if you want to give that a try.

---

> Commented on: 2020-06-08 by @detly

I am looking into this FYI, I just got stumped a bit. But I think `/snap/bin/barrier.barrierc` is the wrong answer, this seems to launch the binary via the `desktop-launch` script. I suspect the correct path to use would be `/snap/barrier/current/usr/bin/barrierc`, and the `/snap/barrier/current/usr/bin/` part of that might be available as a Snapcraft env var... maybe? I'll keep looking.

---

> Commented on: 2020-06-08 by @p12tic

To me it looks like we shouldn't handle snap services using systemd. Snap has its own way to handle daemons, so I suspect that we should use that, otherwise we risk into hitting strange issues. For example, what happens if the snap daemon itself is not launched yet when we try to start barrier directly? What happens if people include a configuration to handle barrier at snap level? Given that snap attempts to be self-contained app distribution method, we should use the methods recommended by them instead of trying to work around.

---

> Commented on: 2020-06-08 by @detly

Okay, so it looks like:

- proof of concept achieved at least: I can get it to work on the login screen and log in
- `/snap/barrier/current/usr/bin/barrier{c,n}` are the paths to use
- if I remove the line (which I added) `Environment=XAUTHORITY=/var/run/lightdm/root/:0`, I get these errors:

   ```text
   Jun 08 23:18:12 fox barrierc[5186]: [2020-06-08T23:18:12] DEBUG: XOpenDisplay(":0")           
   Jun 08 23:18:12 fox barrierc[5186]: No protocol specified                                     
   Jun 08 23:18:12 fox barrierc[5186]: [2020-06-08T23:18:12] WARNING: secondary screen unavailable: unable to open screen  
   ```
- Barrier continues to work fine after logging in ie. the mouse pointer remains on the client screen where it was, connection is continuous
- When I log out, the connection is lost. There is no disconnect message in the server log. In `journalctl` on the client I see (I've tried to keep it relevant):

   ```text
   Jun 08 23:25:22 fox lightdm[3137]: pam_unix(lightdm:session): session closed for user jason   
   Jun 08 23:25:22 fox barrierc[5405]: [2020-06-08T23:25:22] FATAL: X display has unexpectedly di
   sconnected                                                                                    
   Jun 08 23:25:22 fox barrierc[5405]: [2020-06-08T23:25:22] DEBUG: adopting new buffer          
   Jun 08 23:25:22 fox at-spi-bus-launcher[6990]: XIO:  fatal IO error 11 (Resource temporarily u
   navailable) on X server ":0"                                                                  
   Jun 08 23:25:22 fox at-spi-bus-launcher[6990]:       after 41 requests (41 known processed) wi
   th 0 events remaining.                                                                        
   Jun 08 23:25:22 fox systemd[1]: barrierc@192.168.1.231.service: Main process exited, code=exit
   ed, status=1/FAILURE                                                                          
   Jun 08 23:25:22 fox systemd[1]: barrierc@192.168.1.231.service: Failed with result 'exit-code'
   .
   ```

   I assume the above is where LightDM restarts the X server upon logout (fun fact: I'm old enough to remember when hitting <kbd>ctrl</kbd><kbd>alt</kbd><kbd>backspace</kbd> was a widespread habit after logging out due to video driver issues, until display managers all introduced an option to kill X on logout, and lo, it was a great day for everyone). The barrierc service immediately restarts but not successfully:

   ```
   Jun 08 23:25:23 fox systemd[1]: barrierc@192.168.1.231.service: Service hold-off time over, sc
   heduling restart.                                                                             
   Jun 08 23:25:23 fox systemd[1]: barrierc@192.168.1.231.service: Scheduled restart job, restart
    counter is at 1.                                                                             
   -- Subject: Automatic restarting of a unit has been scheduled
   -- Defined-By: systemd
   -- Support: http://www.ubuntu.com/support
   --
   -- Automatic restarting of the unit barrierc@192.168.1.231.service has been scheduled, as the
   result for
   -- the configured Restart= setting for the unit.
   Jun 08 23:25:23 fox systemd[1]: Stopped Barrier Client connected to 192.168.1.231 (Open-source
    KVM software).
   -- Subject: Unit barrierc@192.168.1.231.service has finished shutting down
   -- Defined-By: systemd
   -- Support: http://www.ubuntu.com/support
   --
   -- Unit barrierc@192.168.1.231.service has finished shutting down.
   Jun 08 23:25:23 fox systemd[1]: Starting Barrier Client connected to 192.168.1.231 (Open-sourc
   e KVM software)...
   -- Subject: Unit barrierc@192.168.1.231.service has begun start-up
   -- Defined-By: systemd
   -- Support: http://www.ubuntu.com/support
   --
   -- Unit barrierc@192.168.1.231.service has begun starting up.
   Jun 08 23:25:23 fox systemd[1]: Started Barrier Client connected to 192.168.1.231 (Open-source
    KVM software).
   -- Subject: Unit barrierc@192.168.1.231.service has finished start-up
   -- Defined-By: systemd
   -- Support: http://www.ubuntu.com/support
   --
   -- Unit barrierc@192.168.1.231.service has finished starting up.
   --
   -- The start-up result is RESULT.
   Jun 08 23:25:23 fox barrierc[7006]: [2020-06-08T23:25:23] DEBUG: XOpenDisplay(":0")
   Jun 08 23:25:23 fox barrierc[7006]: [2020-06-08T23:25:23] WARNING: secondary screen unavailabl
   e: unable to open screen
   Jun 08 23:25:23 fox barrierc[7006]: [2020-06-08T23:25:23] DEBUG: retry in 60 seconds
   Jun 08 23:25:23 fox barrierc[7006]: [2020-06-08T23:25:23] DEBUG: event queue is ready
   ```

   Then as the log says, 60 seconds pass and it restarts successfully this time:

   ```
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: XOpenDisplay(":0")           
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: xscreensaver window: 0x000000
   00                                                                                            
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: screen shape: 0,0 1920x1080  
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: window is 0x00600004         
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: adopting new buffer          
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] DEBUG: opened display               
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] NOTE: started client                
   Jun 08 23:26:23 fox barrierc[7006]: [2020-06-08T23:26:23] NOTE: connecting to '192.168.1.231':
    192.168.1.231:24800                                                               
   ```

Personally I'd be happy with this behaviour over not having it at all, but I don't know what you want to do about the `Xauthority` thing. My addition there is a hack, it will work for LightDM but not others, and I don't know how well it translates between distros.

Happy to try more tinkering.

---

> Commented on: 2020-06-08 by @detly

> To me it looks like we shouldn't handle snap services using systemd. Snap has its own way to handle daemons, so I suspect that we should use that, otherwise we risk into hitting strange issues.

I don't disagree, this was the theme of the comments on [my forum question](https://forum.snapcraft.io/t/is-xdg-data-home-modified-by-snaps-adapters/17826). Also note a recent comment regarding the risks of running snap binaries directly.

Most of my notes above might well apply to non-snap distribution too, though. I'll find some time to build and run it from source so I can at least confirm whether it does the right thing in that case.

---

> Commented on: 2020-06-08 by @maxiberta

> I don't disagree, this was the theme of the comments on [my forum question](https://forum.snapcraft.io/t/is-xdg-data-home-modified-by-snaps-adapters/17826). Also note a recent comment regarding the risks of running snap binaries directly.

If I understand correctly, what was suggested in that thread is adding new `app:` entries to [`snap/snapcraft.yaml`](https://github.com/debauchee/barrier/blob/master/snap/snapcraft.yaml) with no `destop-launch` wrapper, so that the binaries can still run _confined via snapd_ with the correct set of paths & libraries pointing to the snap contents instead of the host's, but with no desktop-specific environment.

---

> Commented on: 2020-06-08 by @detly

@maxiberta Yes, but also there's the qualifier <q>if you do not want to re-pack the snap as a service</q> — and I've read elsewhere that snap has specific packaging for daemons distinct from desktop apps.

---

> Commented on: 2020-06-09 by @maxiberta

> @maxiberta Yes, but also there's the qualifier "if you do not want to re-pack the snap as a service" — and I've read elsewhere that snap has specific packaging for daemons distinct from desktop apps.

Just for clarity, adding a service to the snap requires adding a new section below `apps:` to [`snap/snapcraft.yaml`](https://github.com/debauchee/barrier/blob/master/). It might look something [like this](https://github.com/maxiberta/barrier/commit/66e474a7d337501370e2f32f65915173bc4321cc):
```yaml
apps:
  ...
  barriers-daemon:
    command: usr/bin/barriers --enable-crypto --display $DISPLAY --debug $LOG_LEVEL --config $SNAP_DATA/barrier.conf --no-daemon
    environment:
      DISPLAY: :0
      LOG_LEVEL: INFO
    plugs: *plugs
    daemon: simple
  ...
```
This is just an example. It starts the service, but I don't think it does anything useful atm. Config file is at `/var/snap/barrier/current/barrier.conf`. Some extra configuration can be handled [via snap configure hook](https://snapcraft.io/docs/adding-snap-configuration).

Are there plans to support something similar with the flatpak?

---

> Commented on: 2020-06-25 by @murrant

Hello, doesn't it make sense to offer user based scripts?  That way users can start and stop the client/server with the user session.

---

> Commented on: 2020-06-27 by @shymega

> Hello, doesn't it make sense to offer user based scripts? That way users can start and stop the client/server with the user session.

Not sure what you mean. The systemd services remove the need for scripts.

---

> Commented on: 2020-06-27 by @murrant

> Not sure what you mean. The systemd services remove the need for scripts.

Sorry, by script I meant systemd user unit.   Like dbus for example, `systemctl --user status dbus`.  This way you could `systemctl --user enable barrierc@server` and it would connect automatically when you log in (and not when someone else logs in).

---

> Commented on: 2020-07-18 by @monoprosito

So this will be ready soon? It's one of the features I've been looking forward to most for Barrier!

---

> Commented on: 2020-09-21 by @github-actions[bot]

This PR has been automatically marked as stale due to inactivity. It will be closed if no further activity occurs. Thank you for your contributions.

---

> Commented on: 2020-11-15 by @ryukinix

That is a really interesting feature. I have some problems to get client setup working properly automatically at boot time, so a better option I think it would be config as systemd service.

---

> Commented on: 2021-01-10 by @p12tic

@simons-public @detly Is my understanding correct that the only thing that is left in this PR is that Barrier is unavailable for 1 minute after logout? This is an easy thing to change, the timeout is hardcoded.

---

> Commented on: 2021-01-11 by @detly

> Hello, doesn't it make sense to offer user based scripts? That way users can start and stop the client/server with the user session.

The problem is that you can't start the user session.

---

> Commented on: 2021-01-11 by @detly

> @simons-public @detly Is my understanding correct that the only thing that is left in this PR is that Barrier is unavailable for 1 minute after logout? This is an easy thing to change, the timeout is hardcoded.

So the big issue I think is this:

> I don't know what you want to do about the Xauthority thing. My addition there is a hack, it will work for LightDM but not others, and I don't know how well it translates between distros.

This was always a problem, Synergy/Barrier was just about the only use case for "sharing" `XAUTHORITY` between the display manager and another process, so everything I ever made work was a total hack and I never really came up with a reliable method.

---

> Commented on: 2021-01-11 by @murrant

> The problem is that you can't start the user session.

I'll give you that it doesn't make as much sense for the server to run only when logged in (although in some cases that might be desirable).  But the client absolutely makes sense to start/stop with the user session.

---

> Commented on: 2021-01-12 by @detly

> But the client absolutely makes sense to start/stop with the user session.

Oh I see what you're saying. Yes, so far I've been using Gnome's startup settings to launch it; a user systemd script might be less DE-dependent.

---

> Commented on: 2021-01-27 by @fiskhest

> Unrelated: I have also found that I need:
> 
> ```
> Environment=XAUTHORITY=/var/run/lightdm/root/:0
> ```
> 
> ...and the appropriate lines in `/var/lib/snapd/apparmor/profiles/snap.barrier.barrierc` to allow access to that. It seems very fragile eg. what if the user is using GDM or KDM or whatever? But I tried years ago to work out a generic ["where is the Xauthority file" mechanism](https://lists.ubuntu.com/archives/ubuntu-devel/2013-November/037913.html) with no results grimacing.

I might be misunderstanding something here, but is a `xauth info | grep "Authority file" | awk '{print $3}'` helpful for this?

---

> Commented on: 2021-01-27 by @afranke

Have you considered adding a [systemd socket](https://www.freedesktop.org/software/systemd/man/systemd.socket.html)? That would mean that instead of running Barrier from boot, whether they use it or not, one could have it launched on demand, as soon as a client attempts to connect to the port.

---

> Commented on: 2021-01-29 by @detly

> I might be misunderstanding something here, but is a `xauth info | grep "Authority file" | awk '{print $3}'` helpful for this?

@afranke I don't think so, but I haven't tested it. The issue is you need to be able to find the Xauthority file before a user is logged in, as the same user and *in the same session* as the display manager. So eg. running that command as (say) the lightdm user via `su` in a framebuffer console (as I once tried) yielded nothing.

---

> Commented on: 2021-01-29 by @detly

One possibility would be to:
- write scripts for each display manager (I think they all have some sort of "run this on start" config, right?) that runs ~~@afranke's~~ @fiskhest's command and writes the results to a file in eg. `/var` or `/etc/defaults` or somewhere
- barrier's systemd unit can then read that file to get the Xauthority file path

...maybe?

---

> Commented on: 2021-01-29 by @murrant

Keep in mind there are different server use cases.
A. start barrier server before login, and keep the connection going.  Needs to be a system unit.
B. Login as user, then start barrier server.  Stop server after logout  (this is the use case I suggest for user systemd unit)

Client should always be the same only start when logged in, so a user unit makes perfect sense.

---

> Commented on: 2021-05-09 by @Foxboron

These services should be changed to exclude any references to the `DISPLAY` variable and `XAUTHORITY`. They should be installed as user services towards `/usr/lib/systemd/user/`.

I think most of the concerns here is a none issue as any X server sessions should be sourcing these variables into the user slice for the services to use. A script similar to the one found on Arch should be on most distros.
```
λ ~ » cat /etc/X11/xinit/xinitrc.d/50-systemd-user.sh
#!/bin/sh

systemctl --user import-environment DISPLAY XAUTHORITY

if command -v dbus-update-activation-environment >/dev/null 2>&1; then
    dbus-update-activation-environment DISPLAY XAUTHORITY
fi
```

---

> Commented on: 2021-05-10 by @detly

> These services should be changed to exclude any references to the `DISPLAY` variable and `XAUTHORITY`. They should be installed as user services towards `/usr/lib/systemd/user/`.

@Foxboron when do they start if they are installed in this way?

---

> Commented on: 2021-05-10 by @Foxboron

> > These services should be changed to exclude any references to the `DISPLAY` variable and `XAUTHORITY`. They should be installed as user services towards `/usr/lib/systemd/user/`.
> 
> @Foxboron when do they start if they are installed in this way?

You should be capable of specifying `PartOf=graphical-session.target` or something similar to have it start after X. The network target is not given to the user session so there is no way to say "after network has started" and you'd need to rely on retries for that.

---

> Commented on: 2021-05-10 by @detly

> You should be capable of specifying PartOf=graphical-session.target or something similar to have it start after X. The network target is not given to the user session so there is no way to say "after network has started" and you'd need to rely on retries for that.

@Foxboron I'm struggling to find documentation for `graphical-session.target`, but it looks like this is related to desktop environment sessions, and not the display manager. Is that right?

---

> Commented on: 2021-05-10 by @Foxboron

https://www.freedesktop.org/software/systemd/man/systemd.special.html#Special%20Passive%20User%20Units

It should be ran whenever after a graphical session is started. Not sure I understand the question, is it whether or not the target is activated at the login manager stage?

---

> Commented on: 2021-05-10 by @detly

@Foxboron The problem that remains to be solved in this issue is how to run the Barrier server in a way that doesn't need the user to physically (ie. with a connected keyboard) or automatically log in first ie. how can a user *use Barrier* to log in to a headless server and start their graphical session. So previous suggestions of having a systemd user unit don't address this.

Now, all of the examples for `graphical-session.target` seem to involve graphical user login sessions, which makes me think this won't work. But maybe that's just what they chose for the docs, and the target is capable of running under the display manager too?

---

> Commented on: 2021-05-10 by @Foxboron

Then I got the issue. I don't think it's possible.

X needs to know the cookie and display variable. Neither of which you have access too at that stage, you don't even know who the user is going to be. I'm a little bit unsure how the hand-off between the login manager and the X session towards the user work.

I don't think you'll find any solutions here which is going to satisfy people that want cold-boot to work for the barrier server.

---

> Commented on: 2021-05-10 by @detly

> X needs to know the cookie and display variable. Neither of which you have access too at that stage, you don't even know who the user is going to be.

I don't understand this. Surely X *can* know the cookie and display variable for the display manager, because (a) the display manager is running, it must exist and (b) I've written some questionable and DE/DM/distro specific scripts to get that and put it in Barrier's environment. (No one wants to write and maintain thirty or so such scripts for this purpose though, which is why I'd hoped we'd solve it in this issue.) The user is simply whatever user the DM runs as eg. a user named `lightdm`.

If you're wondering about continuing a Barrier session cleanly across the act of logging in, then yes, that's an issue. But (a) having Barrier disconnect immediately after log in is a much smaller issue than not being able to use it at all and (b) Barrier can automatically reconnect anyway.

### DIFF
--- a/dist/systemd/barrierc@.service.in
+++ b/dist/systemd/barrierc@.service.in
@@ -37,7 +37,7 @@ Environment=XDG_DATA_HOME=/var/lib/barrier/barrier@%i
 # TrustedServers.txt Directory
 Environment=FP_DIR=/var/lib/barrier/barrier@%i/barrier/SSL/Fingerprints
 # Ensure the Fingerprints directory exists
-ExecStartPre=+/usr/bin/mkdir -p "${FP_DIR}"
+StateDirectory=barrier/barrier@%i/barrier/SSL/Fingerprints
 
 # This uses openssl commands and grep to get the server's key and 
 # store it in the TrustedServers.txt file. OpenSSL is a requirement
--- a/dist/systemd/barriers.service.in
+++ b/dist/systemd/barriers.service.in
@@ -26,9 +26,8 @@ Environment=DISPLAY=:0
 Environment=XDG_DATA_HOME=/var/lib/barrier/barrier:24800
 # SSL data directory
 Environment=CERT_DIR=/var/lib/barrier/barrier:24800/barrier/SSL
-
-# Create the certificate directory
-ExecStartPre=+/usr/bin/mkdir -p ${CERT_DIR}
+# Ensure the SSL directory exists
+StateDirectory=barrier/barrier:24800/barrier/SSL
 # Create the Barrier.pem certificate if it doesn't exist
 ExecStartPre=/usr/bin/sh -c "[ -f ${CERT_DIR}/Barrier.pem ] || openssl req -x509 -nodes -days 365 -subj '/CN=Barrier' -newkey rsa:2048 -text -keyout ${CERT_DIR}/Barrier.pem -out ${CERT_DIR}/Barrier.pem"
 
--- a/dist/systemd/barriers@.service.in
+++ b/dist/systemd/barriers@.service.in
@@ -38,9 +38,8 @@ Environment=DISPLAY=:0
 Environment=XDG_DATA_HOME=/var/lib/barrier/barrier%i
 # SSL data directory
 Environment=CERT_DIR=/var/lib/barrier/barrier%i/barrier/SSL
-
-# Create the certificate directory
-ExecStartPre=+/usr/bin/mkdir -p ${CERT_DIR}
+# Ensure the SSL directory exists 
+StateDirectory=barrier/barrier%i/barrier/SSL
 # Create the Barrier.pem certificate if it doesn't exist
 ExecStartPre=/usr/bin/sh -c "[ -f ${CERT_DIR}/Barrier.pem ] || openssl req -x509 -nodes -days 365 -subj '/CN=Barrier' -newkey rsa:2048 -text -keyout ${CERT_DIR}/Barrier.pem -out ${CERT_DIR}/Barrier.pem"
 


CMakeLists.txt
```diff
@@ -419,6 +419,15 @@ endif()
...
     configure_files (${CMAKE_CURRENT_SOURCE_DIR}/dist/rpm ${CMAKE_BINARY_DIR}/rpm)
+    if (DEFINED ENV{SNAPCRAFT_PART_BUILD})
+        set (SYSTEMD_EXEC_PATH_PREFIX "/snap/bin/barrier.")
+    else()
+        set (SYSTEMD_EXEC_PATH_PREFIX "/usr/bin/")
+    endif()
+    configure_files (${CMAKE_CURRENT_SOURCE_DIR}/dist/systemd ${CMAKE_BINARY_DIR}/systemd)
+    install(FILES ${CMAKE_BINARY_DIR}/systemd/barrierc@.service DESTINATION lib/systemd/system)
+    install(FILES ${CMAKE_BINARY_DIR}/systemd/barriers@.service DESTINATION lib/systemd/system)
+    install(FILES ${CMAKE_BINARY_DIR}/systemd/barriers.service DESTINATION lib/systemd/system)
```
whot, line 430: No idea how to do this in cmake but `pkg-config --variable systemd_system_unit_dir systemd` is the best way to get this, straight from the horse's mouth. Requires `systemd.pc` to be installed.


But that aside - wouldn't installing this as user service be a better option? To run X and thus barrier you need a logged-in user anyway. And at that point you can enable/disable it per user and the service has access to the user's home directory (and thus config files, etc.). This goes doubly for the barrierc service.

shymega: I would tend to agree, but equally we can't assume `systemd.pc` is installed? I prefer `--user` services, but some people have said they want Barrier to load before login. This just isn't feasible. Personally, I'd rather go for `--user`...

whot: >  we can't assume systemd.pc is installed?

Two options here: either require the dependency (like all others we already require anyway) or default to pkgconfig and if not available pick a fallback path. Since this is unlikely to ever change much anyway, that's a fairly safe option too.

As for user vs system: maybe a configure-time build option?

shymega: Second option feels more sane to me. I think option for user vs. system might not be suited for configure-time - bear in mind other distros, if not already, will be packaging Input Leap, wouldn't that result in more packages for user vs. system approaches? Or am I reading this wrong?

whot: depends mostly on packaging, I guess. But what you could also do is simply install both. pipewire for example does this via the meson options `-Dsystemd-system-service=enabled` (disabled by default) and `-Dsystemd-user-service=enabled` (enabled by default).

Also, just because they're installed doesn't mean they're *enabled* (`systemctl enable --now input-leap` or so) and what is enabled by default is a distribution-specific toggle anyway.

shymega: Yeah, let's just install user services for now, as system services are a bit more hacky, and require the IL config to be stored somewhere, which AFAIK the approach is user-dependent. I can't see system services working for IL just yet. It's something we can explore.

scrummyin: Hey folks, I wanted to suggest writing files to `/etc/xdg/autostart` , because it seems that [arch](https://wiki.archlinux.org/title/XDG_Autostart) and [sway (a wayland i3 clone)](https://github.com/swaywm/sway/issues/1423) push people that way. With a sort of your desktop manager should help manage session autostarting, but if not, just use [dex](https://github.com/jceb/dex) mentality. For background some of the consideration into this suggestion.

It looks like [systemd wasn't really designed for a per seat/per session](https://lists.freedesktop.org/archives/systemd-devel/2014-March/017552.html) service. I stumbled on the above from the Note at the bottom of [Arch's systemd's How it works](https://wiki.archlinux.org/title/systemd/User#how_it_works), which explains that it's more a per user than per session thing.

I see the `Environment=DISPLAY=:0` which in fighting with nvidia and gdm3 today, I discovered that might [not be true for a good amount of environments](https://askubuntu.com/a/1323971).

I considered but did not want to suggest just writing to [.xprofile](https://wiki.archlinux.org/title/Xprofile), as that doesn't seem to fit with upcoming wayland.

P.S. Thanks for all the awesome work so far. I've built the project on my linux boxes and just need to do the same on windows (the depends are a bit bigger) to replace synergy. Also, I'd be happy to whip up a PR if you all are interested in using `.desktop` files.

whot: > I consider systemd/user as a good alternative for dex's autostart functionality and switched to it recently

that's from the [dex README](https://github.com/jceb/dex) itself :) So that's not exactly a huge endorsement of using autostart + dex.

prahal: to me the users complaining about systemd user session not allowing to use IL before login means before login as a user.
The graphical login manager also runs as a user so this graphical login manager could have its own systemd user session service instance for IL and we would have the IL at boot.
The poster seems to not understand the graphical login manager is itself logged in as its own user, only automatically. I do not believe they requested IL to be working before the graphical login manager.


About per seat:
Also, in this patchset, the system-wide systemd service would not help with multi-seat either. Either way, I see no shortcomings in binding IL to the user session instead of the login session.
However, I agree that IL should be seat-aware in that we should have different screen mapping depending on your seat (ie you could have boxes with one display on one seat, three on another, two with two displays, etc). IL config syntax does not support multi-seat as far as I know.

In https://lists.freedesktop.org/archives/systemd-devel/2014-March/017552.html the poster confused login session to seat (he says about login session then says he meant different window managers, so I believe he meant different graphical logins). Lennart discards his request because in his understanding graphical display managers will be per user so he sees no point in the request (as systemd already has per-user session support). It is even more obvious in https://lists.freedesktop.org/archives/systemd-devel/2014-March/017543.html where Lennart says that the per login management should be the window manager job, not systemd one. Also, from [this email](https://lists.freedesktop.org/archives/systemd-devel/2014-March/017552.html), I understand that Lennart sees the per-user session display manager has the arbitrator for the devices (including virtual devices like IL input I assume).

This to me means if we want per seat, IL could become part of the display manager or kind of a plugin to it.
Or maybe we could have code in the display manager that spawns systemd user per instance services (one for each seat). That is have one IL server and clients set for each seat on the boxes.
Note that systemd-logind is aware of the seat setup. But as far as I know, we cannot have it spawn programs on login based on seat.

Maybe a bug report for multi-seat support should be opened in input-leap. I believe this to be a feature request which should not prevent this pull request from merging.
The issue already exists with current code. And would not be made worse by systemd user services for IL.

Issue #347, opened 2021-11-04
How to launch client through systray


ghost:
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/347
> 
> Issue created on: 2019-07-01 by @ghost
> Issue last updated on: 2019-08-14

Use case: 
1) Autostart barrier tray icon
2) Have script running and poll for event:
a) Machine plugged in to docking station - if barrier client not started, start it
b) Machine unplugged from docking station - if barrier client started, kill it

Currently this is possible, however without monitoring from GUI. GUI client runs completely separately and is unaware of a second instance of barrier client. This leads to two problems:

1) Killing barrier client, the child proccess of barrier GUI, results in barrier GUI restarting the process. 

2) Spawning a new barrier client does not get registered with the barrier GUI, which is unaware of this instance.

Ideally, the GUI would allow some way to emulate the user action of selecting start/stop service from the tray icon. Unix domain sockets? TCP sockets? DBUS? Something to that effect.

I am aware that automation can be achieved with the barrierc client alone, however the tray icon application cannot be presently made to work with this scenario.

If the devs specify which IPC method would be suitable for this task, if any, I might do a POC. 

---

> Commented on: 2019-08-14 by @shymega

With regards to a IPC method, I would, personally, go for something efficient and portable.

We _could_ use Unix domain sockets for the *nix targets, including macOS. With Windows, there's a IPC method [called][sockets] 'Windows Sockets', which might be suitable. 

[sockets]: https://docs.microsoft.com/en-us/windows/win32/ipc/interprocess-communications#using-windows-sockets-for-ipc

PR #758, opened 2021-11-04
send command to hdmi switch

ghost:
> This PR has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/pull/758
> 
> PR created on: 2020-06-19 by @VE3NEA
> PR last updated on: 2021-04-28

P.O.C. only, please do not merge.

This is the code mentioned in [Issue #347](https://github.com/debauchee/barrier/issues/757) that
sends a source selection command to the HDMI switch.

---

> Commented on: 2020-06-19 by @p12tic

I think it would make sense to have a way to specify arbitrary command that would be run on screen switch. Then users could support this use case without touching Barrier itself.

---

> Commented on: 2020-06-20 by @VE3NEA

An arbitrary command will work, but it will be slower than the proposed solution. The code in this PR keeps the COM port open to reduce the response time.

---

> Commented on: 2020-06-20 by @p12tic

>  The code in this PR keeps the COM port open to reduce the response time.

Could we measure how much time does port opening take? A delay like 100ms would be almost unnoticeable and given how much computation can be done in that time it's hard to believe that opening of the COM port would take more.

---

> Commented on: 2020-06-20 by @VE3NEA

On my system, with a usb-to-rs232 adapter, it takes the port almost a second to open, that's why I re-wrote the code to keep the port open.

---

> Commented on: 2020-06-21 by @shymega

I do think that this is a rather niche use case. I'm not convinced it needs to be in the main Barrier codebase - from what I understand, I can see it will only work on one HDMI switch.

---

> Commented on: 2020-07-14 by @axelson

@VE3NEA couldn't you have another program always running that keeps the port running, and then from barrier you could send a command to your program to tell it to switch?

---

> Commented on: 2020-07-14 by @VE3NEA

I wrote that code when I was still waiting for the switch to arrive. Now I have it and, unfortunately, it takes the switch/monitor/videocard 2 to 5 seconds to start showing the other screen. 0.7 seconds spent on opening the port does not make much of a difference, so it makes no sense to keep the port open. @p12tic 's suggestion will work fine.

---

> Commented on: 2020-07-14 by @shymega

An approach which would be more extensible would be to have Barrier support 'hooks'. That way if screen X is activated, Barrier could act on a hook associated with that event to switch the HDMI selection switch. Of course, that's more work...

---

> Commented on: 2020-07-19 by @p12tic

> On my system, with a usb-to-rs232 adapter, it takes the port almost a second to open, that's why I re-wrote the code to keep the port open.

From the architecture point of view, it would be best to implement this "keep port open" functionality as a separate daemon-like program.

Barrier would just call helper programs that would inform daemon that something needs to be sent to the interface. This way the responsibilities are properly separated - barrier does not include any HDMI-related things and the daemon does not include any Barrier-related things.

---

> Commented on: 2020-08-18 by @missmah

👍 This definitely seems like the right direction (hooks on i.e. screen switch or keystroke) and external daemons to handle things like communication to specific HDMI/DP switches.

It would be neat to see what this looks like at the limit. e.g. Having 6 DP monitors and a DP matrix switch, and being able to setup mappings for various work configurations (e.g. one PC use all 6 displays, split them up between 6 PCs, or some combination of the two for different scenarios).

One important question then becomes how does this interact with the Barrier "Configure Server.." map -- which is currently "static"

In this world where you might be changing the PC:Monitor mappings dynamically, it would be nice to be able to also swap "Server Configurations" as part of the same keystroke.

There may be a subset of people who would be happy to swap *all* the displays between different computers (all on A. vs all on B.), and that use case wouldn't require "dynamic server configurations"; but, personally, my use cases often revolve around seeing some subset of my displays persistently, and I don't envision that changing even if I were to add an HDMI/DP switch to dynamically change which PC some displays point at.

---

> Commented on: 2020-08-27 by @maxlock

You don't even need display switching hardware for this to be useful. I'd like to have my monitor switch inputs based upon which host has focus. I can do this using ddc over the video connection itself. For example on a linux platform I could have barrier call ddcutil.

---

> Commented on: 2021-04-28 by @theLMGN

Are hooks implemented in Barrier yet? If so, is there any documentation?

### DIFF
--- a/src/lib/barrier/HdmiSwitch.cpp
+++ b/src/lib/barrier/HdmiSwitch.cpp
@@ -1,0 +1,106 @@
+
+#include "HdmiSwitch.h"
+#include "base/Log.h"
+
+static string lastPortSettings("");
+string deviceName;
+int baudRate, dataBits, parity, stopBits;
+static HANDLE comPort = INVALID_HANDLE_VALUE;
+
+void HdmiSwitch::SendCommand(const string & command, const string & portSettings)
+{
+  try
+  {
+    LOG((CLOG_INFO "Sending command to HDMI switch. Command: %s, portSettings: %s", command.c_str(), portSettings.c_str()));
+    if (comPort == INVALID_HANDLE_VALUE || lastPortSettings.compare(portSettings) != 0)
+    {
+      ParsePortSettings(portSettings);
+      OpenComPort();
+    }
+    SendStringToComPort(command);
+  }
+  catch (std::exception & e)
+  {
+    LOG((CLOG_ERR "Unable to send command to HDMI switch: %s", e.what()));
+  }
+}
+
+void HdmiSwitch::ParsePortSettings(const string& portSettings)
+{
+  lastPortSettings = portSettings;
+  if (comPort != INVALID_HANDLE_VALUE) CloseHandle(comPort);
+
+  try
+  {
+    //device name
+    int posB = 0;
+    int posE = portSettings.find('-');
+    string deviceName = string("\\\\.\\") + portSettings.substr(posB, posE - posB);
+
+    //baud rate
+    posB = posE + 1;
+    posE = portSettings.find('-', posB);
+    baudRate = std::stoi(portSettings.substr(posB, posE - posB));
+
+    //data bits
+    posB = posE + 1;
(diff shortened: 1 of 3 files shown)


xpufx:
What is the current status of this feature?

PR #1418, opened 2021-11-16
Windows build warnings

albertony:
Fixed all compiler warnings when building only the command line programs with Windows with Visual Studio 2022 (VC 14.30):
```
"BARRIER_BUILD_GUI": false,
"BARRIER_BUILD_INSTALLER": false
```

I'm quite sure the added numeric type casts are safe with regards to overflow.

The most controversial change is probably the added header `src/lib/common/win32/winapi.h` for just including `Windows.h` - but with preprocessor directives: `WIN32_LEAN_AND_MEAN`, `STRICT` and `NOMINMAX`. Don't know if you think this is a good idea? (And if you have a better name for it...) Main reason is I needed the `NOMINMAX` to avoid conflict with legacy max macro when using `numeric_limits<int>::max()`. There was (and still is outside of the command line "world" I am compiling atm) quite a few instances of the following:

```
#define WIN32_LEAN_AND_MEAN
#include <Windows.h>
```

An alternative could be to define them at CMake level with `add_compile_definitions`, but would then have to be done for every target..

PS: Congrats on your new project. Looking forward to follow its development!

## Contributor Checklist:

### DIFF
--- a/src/lib/common/win32/winapi.h
+++ b/src/lib/common/win32/winapi.h
@@ -15,14 +15,21 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifndef BARRIER_LIB_COMMON_WIN32_ENCODING_UTILITIES_H
-#define BARRIER_LIB_COMMON_WIN32_ENCODING_UTILITIES_H
+#ifndef BARRIER_LIB_COMMON_WIN32_WINAPI_H
+#define BARRIER_LIB_COMMON_WIN32_WINAPI_H
 
-#include "winapi.h"
-#include <string>
-#include <vector>
-
-std::string win_wchar_to_utf8(const WCHAR* utfStr);
-std::vector<WCHAR> utf8_to_win_char(const std::string& str);
-
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
 #endif
+
+#ifndef STRICT
+#define STRICT
+#endif
+
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
+#include <Windows.h>
+
+#endif // BARRIER_LIB_COMMON_WIN32_WINAPI_H
--- a/src/lib/net/SecureUtils.cpp
+++ b/src/lib/net/SecureUtils.cpp
@@ -245,7 +245,9 @@ std::string create_fingerprint_randomart(const std::vector<std::uint8_t>& dgst_r
     std::size_t i;
     std::uint32_t b;
     int	 x, y;
-    std::size_t len = strlen(augmentation_string) - 1;
+
+    // avoid compiler warning when comparing len to items in field array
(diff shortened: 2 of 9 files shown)


src/lib/common/win32/winapi.h
```diff
@@ -0,0 +1,32 @@
...
+    This package is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
```
p12tic, line 18: Let's use macro-based include guards.

src/lib/net/SecureUtils.cpp
```diff
@@ -241,12 +241,11 @@ std::string create_fingerprint_randomart(const std::vector<std::uint8_t>& dgst_r
      * intersects with itself.  Matter of taste.
      */
     const char* augmentation_string = " .o+=*BOX@%&#/^SE";
-    char *p;
```
p12tic, line 244: Separate commit.

src/lib/net/SecureUtils.cpp
```diff
@@ -241,12 +241,11 @@ std::string create_fingerprint_randomart(const std::vector<std::uint8_t>& dgst_r
...
     std::uint8_t field[FLDSIZE_X][FLDSIZE_Y];
     std::size_t i;
     std::uint32_t b;
     int	 x, y;
-    std::size_t len = strlen(augmentation_string) - 1;
+    std::uint8_t len = static_cast<uint8_t>(strlen(augmentation_string) - 1);
```
p12tic, line 248: This one is not obvious. Add a comment above, something like "// avoid compiler warning when comparing len to items in field array"

src/lib/platform/MSWindowsScreen.cpp
```diff
@@ -52,6 +52,7 @@
 #if defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable: 4706) // assignment within conditional
+#pragma warning(disable: 4996) // GetVersionExA was declared deprecated
```
p12tic, line 55: Separate commit.

p12tic (changes requested):
Thanks a lot for this pull request, warnings are annoying and fixing them is much appreciated.

I agree with everything you've done, it looks great. I'd like the changes to be split into separate commits, like for windows.h header, for the implicit conversion fixes and so on.

albertony:
Sure. Done. Pushed.

p12tic (approved):
Thanks!

p12tic:
Looks like the `mojave-Release` build has trouble launching lately, maybe some kind of abuse prevention is going on as this project currently does not have much track record of not abusing CI.

I'll merge without CI rerun.

PR #1419, opened 2021-11-17
azure-pipelines: Disable macOS mojave build

p12tic:
Azure will drop support for Mojave builds soon and builds are failing due to agent provisioning issues already.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -91,10 +91,6 @@ jobs:
           imageName: "macOS-10.15"
           B_BUILD_TYPE: Release
           BARRIER_VERSION_STAGE: Release
-      mojave-Release:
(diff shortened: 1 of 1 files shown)


shymega:
I think it's best I remove Mojave from the macOS CI PR, which keeps it all together. Alternatively, we could merge this PR, but I'd have to adjust my PR, making more work for updating the Azure Pipelines config...

p12tic:
It's best to keep separate PRs for different purposes. Right now it's not clear when 1385 is going to be merged, however we need this fix right now. I don't think rebasing on top of latest master will be an issue, in the worst case it's just a single merge conflict.

shymega:
I've just squashed #1385, and removed Mojave from the matrix. PR #1385 is now ready for your review.

PR #1421, opened 2021-11-19
Fixed missing free of SSL certificate

albertony:
Fixed missing free of SSL certificate and combining two similar certifcation check functions into one.

I noticed the existing function `verify_cert_fingerprint` called `SSL_get_peer_certificate` without a following `X509_free`. That surely must lead to a resource leak? Also the existing code called helper function `verify_cert_fingerprint` before `ensure_peer_certificate`, but if the peer has no cert (cert is NULL) then `verify_cert_fingerprint` would return false and the `ensure_peer_certificate` not be called. So the only remaining use of `ensure_peer_certificate` was only when there actually was a certificate, to write a log line with the certificate's subject. Therefore I combined these two into one common function, then only one call to `SSL_get_peer_certificate`, followed by a call to `X509_free`, is needed to do both things.

## Contributor Checklist:

* [X] I have created a file in the `doc/newsfragments` directory *IF* it is a

### DIFF
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -25,6 +25,7 @@
 #include "arch/XArch.h"
 #include "base/Log.h"
 #include "base/String.h"
+#include "base/finally.h"
 #include "common/DataDirectories.h"
 #include "io/filesystem.h"
 #include "net/FingerprintDatabase.h"
@@ -705,8 +706,14 @@ bool SecureSocket::verify_cert_fingerprint(const barrier::fs::path& fingerprint_
 
     // calculate received certificate fingerprint
     barrier::FingerprintData fingerprint_sha1, fingerprint_sha256;
+
+    auto* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
+    if (cert == nullptr) {
+        LOG((CLOG_NOTE "peer has no ssl certificate"));
+        return false;
+    }
+    auto cert_free = barrier::finally([cert]() { X509_free(cert); });
     try {
-        auto* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
         fingerprint_sha1 = barrier::get_ssl_cert_fingerprint(cert,
                                                              barrier::FingerprintType::SHA1);
         fingerprint_sha256 = barrier::get_ssl_cert_fingerprint(cert,
@@ -715,6 +722,7 @@ bool SecureSocket::verify_cert_fingerprint(const barrier::fs::path& fingerprint_
         LOG((CLOG_ERR "%s", e.what()));
         return false;
     }
+
 
     // note: the GUI parses the following two lines of logs, don't change unnecessarily
     LOG((CLOG_NOTE "peer fingerprint (SHA1): %s (SHA256): %s",
--- a/src/lib/net/SecureSocket.cpp
+++ b/src/lib/net/SecureSocket.cpp
@@ -461,17 +461,12 @@ SecureSocket::secureAccept(int socket)
     // If not fatal and no retry, state is good
     if (secure_accept_retry_ == 0) {
         if (security_level_ == ConnectionSecurityLevel::ENCRYPTED_AUTHENTICATED) {
-            if (verify_cert_fingerprint(
+            if (verify_peer_certificate(
                         barrier::DataDirectories::trusted_clients_ssl_fingerprints_path())) {
                 LOG((CLOG_INFO "accepted secure socket"));
-                if (!ensure_peer_certificate()) {
-                    secure_accept_retry_ = 0;
-                    disconnect();
-                    return -1;// Cert fail, error
-                }
             }
             else {
-                LOG((CLOG_ERR "failed to verify server certificate fingerprint"));
+                LOG((CLOG_ERR "failed to verify client certificate fingerprint"));
                 secure_accept_retry_ = 0;
                 disconnect();
                 return -1; // Fingerprint failed, error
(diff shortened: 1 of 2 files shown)


src/lib/net/SecureSocket.cpp
```diff
@@ -699,19 +667,30 @@ SecureSocket::disconnect()
...
     // calculate received certificate fingerprint
     barrier::FingerprintData fingerprint_sha1, fingerprint_sha256;
     try {
-        auto* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
         fingerprint_sha1 = barrier::get_ssl_cert_fingerprint(cert,
-                                                             barrier::FingerprintType::SHA1);
+                                                                barrier::FingerprintType::SHA1);
         fingerprint_sha256 = barrier::get_ssl_cert_fingerprint(cert,
-                                                               barrier::FingerprintType::SHA256);
+                                                                barrier::FingerprintType::SHA256);
```
p12tic, line 690: Weird formatting change.

src/lib/net/SecureSocket.cpp
```diff
@@ -699,19 +667,30 @@ SecureSocket::disconnect()
...
+    LOG((CLOG_INFO "peer ssl certificate info: %s", line));
+    OPENSSL_free(line);
+
     // calculate received certificate fingerprint
     barrier::FingerprintData fingerprint_sha1, fingerprint_sha256;
     try {
-        auto* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
         fingerprint_sha1 = barrier::get_ssl_cert_fingerprint(cert,
-                                                             barrier::FingerprintType::SHA1);
+                                                                barrier::FingerprintType::SHA1);
```
p12tic, line 688: Weird formatting change.

src/lib/net/SecureSocket.cpp
```diff
@@ -699,19 +667,30 @@ SecureSocket::disconnect()
...
 }
 
-bool SecureSocket::verify_cert_fingerprint(const barrier::fs::path& fingerprint_db_path)
+bool SecureSocket::verify_peer_certificate(const barrier::fs::path& fingerprint_db_path)
 {
     // ssl_mutex_ is assumed to be acquired
 
+    // ensure peer presented a certificate
+    X509* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
+    if (cert == NULL) {
```
p12tic, line 676: nullptr

src/lib/net/SecureSocket.cpp
```diff
@@ -699,19 +667,30 @@ SecureSocket::disconnect()
...
     barrier::FingerprintData fingerprint_sha1, fingerprint_sha256;
     try {
-        auto* cert = SSL_get_peer_certificate(m_ssl->m_ssl);
         fingerprint_sha1 = barrier::get_ssl_cert_fingerprint(cert,
-                                                             barrier::FingerprintType::SHA1);
+                                                                barrier::FingerprintType::SHA1);
         fingerprint_sha256 = barrier::get_ssl_cert_fingerprint(cert,
-                                                               barrier::FingerprintType::SHA256);
+                                                                barrier::FingerprintType::SHA256);
+        X509_free(cert);
```
p12tic, line 691: Instead of trying to add X509_free to all paths, it's better to automate this using RAII.

Add `auto cert_free = barrier::finally([cert]() { X509_free(cert); });` above immediately below when we know that cert needs to be freed.

p12tic (changes requested):
Thanks for the PR, it looks good in principle. Aside from the comments I added, could you split the fix for the memory leak in a separate commit and the add refactoring to the other. This will make it clear to anyone reading the code in the future where the fix was needed.

p12tic:
Also, could you prefix commit subject with `lib/net: ` and don't exceed the standard limit of 72 characters in the subject line? (Just in case you're interested why, this covers why writing good git commits is important https://chris.beams.io/posts/git-commit/ explains well why git commit)

src/lib/net/SecureSocket.cpp
albertony: Ah, visual studio being "helpfull"..

src/lib/net/SecureSocket.cpp
albertony: Sure. I was trying to follow the majority of other parts of the file.. Though there are both NULL and nullptr used. I must say it is not all that consistent code style in existing code, while you have rather specific expectations for new code, which is kind of hard to know in advance..

src/lib/net/SecureSocket.cpp
albertony: Sure. I've seen that barrier::finally, and understand its use. It just felt unnecessary in this case, with only two known, and very close branches.

src/lib/net/SecureSocket.cpp
p12tic: > while you have rather specific expectations for new code, which is kind of hard to know in advance..

I'm not blaming you, it's impossible to know all expectations. I'm just pointing out the better option, let's fix it efficiently and move on :-)

src/lib/net/SecureSocket.cpp
p12tic: In the future code may evolve and especially in the presence of exceptions it's rather easy to introduce memory leaks.

p12tic: I agree that for the current code `finally` was probably not needed, I'm more focusing in how this could break in the future. In the long term it would be great if all such allocations could be freed by `finally`, then the chance of resource leaks would be greatly reduced.

albertony:
Pushed, as two commits this time. Hope I covered all your comments.

p12tic (approved):
Looks good, thanks!

PR #1429, opened 2022-01-06
clean_build: Use cmake --build . --parallel

p12tic:
This ensures that the all CPU cores are used without introducing any non-portable scripting.

### DIFF
--- a/clean_build.sh
+++ b/clean_build.sh
@@ -36,5 +36,5 @@ mkdir build || exit 1
 cd build || exit 1
 echo "Starting Barrier $B_BUILD_TYPE build..."
 "$B_CMAKE" "$B_CMAKE_FLAGS" .. || exit 1
-make || exit 1
+"$B_CMAKE" --build . --parallel || exit 1
 echo "Build completed successfully"


shymega (approved):
Looks fine to me. Liking the way it doesn't rely on `make` now.

shymega:
Actually, just one thing - what compatibility does this change impose on older versions of CMake? That's my one concern, other than that I have no objections.

p12tic:
Our requirements are CMake 3.4 (see CMakeLists.txt), which according to the docs already had --build command.

albertony:
But not --parallel until [3.12](https://cmake.org/cmake/help/v3.12/release/3.12.html).

shymega:
> But not --parallel until [3.12](https://cmake.org/cmake/help/v3.12/release/3.12.html).

Yeah, that was my concern too.

albertony:
3.12 came out july 2018, so bumping minimum requirement to that should not be too controversial...

p12tic:
Hm, indeed, I should have checked --parallel too.

Though this is not actually real minimum requirement on CMake version. clean_build.sh is just a convenience script.

shymega:
That's true, it's only in the script. For package maintainers, I don't see much of an issue -- they have their own processes, which are usually quite robust -- and bumping the requirement as @albertony rightly points out won't be too much trouble.

PR #1443, opened 2022-02-12
Run find/replace on translations

shymega:
This is a bit hacky, but I ran `sed` on the translations, to change
'Barrier' to 'Input Leap', etc.

It seems to compile fine, but I haven't yet tested translations, as I'm
only fluent in English, unfortunately.

There are still references to `barriers` and `barrierc`, but we haven't
renamed those yet, so this will need more work.

Signed-off-by: Dom Rodriguez <shymega@shymega.org.uk>

## Contributor Checklist:


shymega:
Actually, I'm going to convert this to a draft. That way, we can continue the work on renaming `barriers` and `barrierc` - if you disagree @p12tic, we can revert that conversion.

p12tic:
@shymega barriers and barrierc conversion will need more work because we want to migrate the configuration automatically. I think let's limit this PR to translations only.

p12tic:
@shymega PR LGTM.

I think commit title is a bit unclear right now: "Run find/replace on translations" does not say what we are replacing. Something like "Rename Barrier to Input Leap in translations" would be better.

p12tic (approved):

shymega:
Just renamed.

Issue #1441, opened 2022-02-11
Look into new CI


shymega:
This issue is to track the transition into a new CI platform, or to be
self-hosted, that can test older versions of macOS, Windows and Linux, for those
using older OSes.

The current CI is Azure Pipelines, which is soon discontinuing Windows Server
2016, which would mean our current CI configuration would break for Windows
releases.

AP also doesn't have Windows 11 as a build agent, or older Windows releases,
meaning new releases could break functionality on these older OSes. The same
goes for macOS and Linux.

a5ehren:
I'm poking at this to learn GitHub Actions - is there some matrix of OS Support that you want to be able to test builds on for CI use?

Github Actions seem to make it pretty easy to spawn and use docker containers inside the ubuntu-latest image, so that will at least cover basically any Linux builds you want. 

I will admit to being completely ignorant of how building on MacOS and Windows works, though. I suspect that the images available in Actions for MacOS (10.15 and 11 for now, 12 coming soon) will cover the vast majority of users, but I have no clue what to do about Windows coverage.

shymega:
Hi. I've been away from the project for a while, but I've tracked the PR discussion, and I really must stress that using two CI platforms isn't desirable for me. I'd rather have one. Azure is generally stable, the only issue is the agents. I know using GH Actions and Azure was discussed, but let's keep it KISS.

I don't really like the idea of more CI platforms to maintain. With regards to your PRs, let's discuss them here first, so we can iron out issues, think about the approach etc.

luser:
Options for macOS cloud compute are thin on the ground, mostly because of Apple's draconian virtualization restrictions. GitHub Actions does have [an option to self-host runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners), so you could certainly acquire some mac minis and use them as runners for mac testing.

shymega:
I wanted to come back to this. We now have GH Actions AND Azure Pipelines. They are both run on the same infrastructure, so I think we should move completely to GH Actions, and keep it simple. What's the point of two CIs? It's unnecessary. More complexity. I know @p12tic said to me he'd happily maintain both, but it's just such a waste of compute cycles, and adds very little to the project.

I want to move to one CI platform, and if GH Actions supports self-hosted runners, we could test on M1, and Windows 11.

Let's do this!

p12tic:
[...]

PR #1465, opened 2022-04-15
Issue #1441 Make changes for Github Actions CI and Innosetup 6.2.0

a5ehren:
This PR is what it takes to move the existing Azure Pipelines CI to Github Actions and use Innosetup 6.2. Not presenting this as a fully done solution, it still needs work and input.

Other changes:
- Bump Qt for Windows builds to 5.15.2 to match MacOS and Linux
- Use VS2022 and the latest VC++ redist for Windows
- Combination of the above two changes and Innosetup 6.2 security is that minimum Windows version is now Win7 SP1
- Tell the MacOS build to target 10.15 instead of 10.9
- Replace the custom dependency fetcher for Innosetup with [the community-standard script](https://github.com/DomGries/InnoDependencyInstaller)

Still needs work:
- Build output uploads need some tweaking to work in a form we would actually expect
- The ISS script needs a lot more updates to finish the branding move to Input Leap
- The Github Action needs to run on other Linux distros via containers (Arch and Fedora at least?)
- Need to figure out support matrix for Windows and MacOS and what that looks like for building vs running


p12tic:
Thanks for the PR.

The problem with Azure Pipelines is just that it provides too small coverage in terms of available platforms. Ideally we want InputLeap to run on ancient hardware, this means supporting older software platforms too, which is usually not hard, just needs some `#ifdefs` here and there. Github Actions does not improve the situation much in this regard, at least it will not meet our full requirements. Having said that I think it makes sense to have more than one CI available because different platforms usually support slightly different sets of non-Linux OSes.

The plan was to have a private Buildbot or Jenkins-based CI for all the obsolete platforms around.

I think this PR is good in principle, let's split code changes into multiple PRs so that if anything causes a regression it's easier to revert them. Let's not remove Azure Pipelines configuration, instead let's reuse as much code as possible between Azure Pipelines and Github CI.

More precise and actionable comments follow:

 - Let's move "Bump Qt for Windows builds to 5.15.2 to match MacOS and Linux" to separate PR.
 - Let's move "Use VS2022 and the latest VC++ redist for Windows" to separate PR.
 - What exactly brings the version requirement up to Win7 SP1? We may want to support older Windows versions, so we may want an option to build for older releases.
 - We can't bump macOS target to newer versions because this will exclude many people with older hardware. InputLeap is very useful in these scenarios when there are primary and secondary workstations and the secondary can be crap old machine.
 - Please move "Replace the custom dependency fetcher for Innosetup with [the community-standard script](https://github.com/DomGries/InnoDependencyInstaller)" to separate PR.
 - Let's not delete Azure Pipelines CI. This way if github screws up we will still have CI coverage. It requires minimal oversight anyway.

a5ehren:
If I drop back to VS2019 and use the last VC++ redist for that, we get WinXP back. I don't know what MS did in the later VC++ to break XP.

I also think I can set a flag in the ISS file to let the installer work in XP as well. Innosetup says they bumped their default minimum to 7SP1 to prevent a DLL hijacking attack.

I'll split the changes like you suggested as well.

a5ehren:
Ok...dealing with old versions of Windows in a CI context is a nightmare. My build experience is all with Linux, so I didn't realize this was all such a pain.

I'm just going to leave out my Windows changes and make a PR with the Actions for MacOS and Linux.

a5ehren:
These are the lines from Innosetup 6/6.1 talking about why they dropped pre-7 support:

> OS requirements change: Windows 2000, XP, and Server 2003 are no longer supported. Windows Vista is the minimum supported operating system.
> 
> Change in default behavior: Starting with Inno Setup 6.1 the [Setup] section directive MinVersion defaults to 6.1sp1, so by default Setup will not run on Windows Vista or on versions of Windows 7 and Windows Server 2008 R2 which have not been updated. Setting MinVersion to 6.0 to allow Setup to run on Windows Vista is supported but not recommended: Windows Vista doesn't support some of Setup's security measures against potential DLL preloading attacks so these have to be removed by the compiler if MinVersion is below 6.1 making your installer less secure on all versions of Windows.

PR #1467, opened 2022-04-15
Issue #1441 Update the windows builders to use Qt 5.15.2

a5ehren:
to match MacOS and Linux. This is the last OSS version of Qt 5.x.

### DIFF
--- a/azure-pipelines/download_install_qt.ps1
+++ b/azure-pipelines/download_install_qt.ps1
@@ -1,7 +1,7 @@
 $ErrorActionPreference = "Stop"
 
 $qli_install_version = '2019.05.26.1'
-$qt_version = '5.13.0'
+$qt_version = '5.15.2'
 
 New-Item -Force -ItemType Directory -Path ".\deps\"
 
@@ -21,5 +21,5 @@ Write-Output 'Installed QLI Installer Dependencies'
 Write-Output 'Starting QT Installer'
 $Env:QLI_OUT_DIR = ".\deps\Qt\Qt$qt_version"
 $Env:QLI_BASE_URL = "https://download.qt.io/online/qtsdkrepository/"
-python .\deps\qli-installer\qli-installer.py $qt_version windows desktop win64_msvc2017_64
+python .\deps\qli-installer\qli-installer.py $qt_version windows desktop win64_msvc2019_64
 Write-Output 'Installed QT Installer'
--- a/azure-pipelines/build_env_tmpl.bat
+++ b/azure-pipelines/build_env_tmpl.bat
@@ -1,6 +1,5 @@
 set B_BUILD_TYPE=%CI_ENV_BUILD_TYPE%
 set B_QT_ROOT=%cd%\deps\Qt
-set B_QT_VER=Qt5.13.0\5.13.0
-set B_QT_MSVC=msvc2017_64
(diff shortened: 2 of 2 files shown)


azure-pipelines/download_install_qt.ps1
```diff
@@ -21,5 +21,5 @@ Write-Output 'Installed QLI Installer Dependencies'
 Write-Output 'Starting QT Installer'
 $Env:QLI_OUT_DIR = ".\deps\Qt\Qt$qt_version"
 $Env:QLI_BASE_URL = "https://download.qt.io/online/qtsdkrepository/"
-python .\deps\qli-installer\qli-installer.py $qt_version windows desktop win64_msvc2017_64
+python .\deps\qli-installer\qli-installer.py $qt_version windows desktop win64_msvc2019_64
```
p12tic, line 24: Is this needed for Qt version change?

p12tic:
Could you adjust the commit message to conform to the best practices? First line should be short description of at most 72 characters (otherwise tools may start hiding part of the message), and any additional information afterwards. Issue number is not needed.

E.g.:

```
Update the windows builders to use Qt 5.15.2

This matches MacOS and Linux versions and is the last OSS version of Qt 5.x.
```

(https://initialcommit.com/blog/git-commit-messages-best-practices is a good resource about best practices for commit messages if you think reading about this may be useful).

azure-pipelines/download_install_qt.ps1
a5ehren: This comes from upstream. For whatever reason 5.15.0 has the msvc_2017 version but 5.15.1+ do not. 

The msvc2019 change passed build on the Azure VS2017 builder though.

shymega:
> Could you adjust the commit message to conform to the best practices? First line should be short description of at most 72 characters (otherwise tools may start hiding part of the message), and any additional information afterwards. Issue number is not needed.
> 
> E.g.:
> 
> ```
> Update the windows builders to use Qt 5.15.2
> 
> This matches MacOS and Linux versions and is the last OSS version of Qt 5.x.
> ```
> 
> (https://initialcommit.com/blog/git-commit-messages-best-practices is a good resource about best practices for commit messages if you think reading about this may be useful).

Whilst I agree with most of your commit suggestions, I strongly disagree that the issue number is 'not needed'. It's actually really useful, and I would encourage @a5ehren to amend their commit to include the relevant issue in the commit body. I imagine you were referring to the first line, which arguably might not need the issue #, but there's simply no harm in it. If anything,given the tight Git integration, we have a more streamlined way of tracking bugs and features!

azure-pipelines/download_install_qt.ps1
shymega: The only thing I see with this is that we drop support for older platforms. I'd like to see conditionals.

a5ehren: Conditionals are tough since your release builds only come from one place. I can roll it back to 5.15.0 if you're nervous about the msvc2019 part.

p12tic (approved):

p12tic:
> I imagine you were referring to the first line, which arguably might not need the issue #, but there's simply no harm in it.

Agreed. I added issue number when rebasing.

PR #1469, opened 2022-04-22
Qt6

jpwhiting:
## Contributor Checklist:

Not sure when but when wanting to port to Qt 6 feel free to use this as a starting point. With these 2 commits I'm able to build here on macos (though the gui hangs when clicking Start and a couple warnings are in the terminal when running directly. I'll look into those next.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -1,6 +1,14 @@
 cmake_minimum_required (VERSION 3.4)
 
-find_package (Qt5 REQUIRED COMPONENTS Core Widgets Network LinguistTools)
+find_package (QT NAMES Qt6 Qt5 REQUIRED)
+find_package (Qt${QT_VERSION_MAJOR} COMPONENTS Core Widgets Network LinguistTools REQUIRED)
+# If Qt6 is in use, also link to Core5Compat
+if (Qt6_FOUND)
+    find_package (Qt6 REQUIRED COMPONENTS Core5Compat)
+endif()
+
+message(STATUS "Using Qt" ${QT_VERSION_MAJOR})
+
 
 set (CMAKE_AUTOMOC ON)
 set (CMAKE_AUTORCC ON)
@@ -165,7 +173,11 @@ set(TS_FILES
 )
 
 set_source_files_properties(${TS_FILES} PROPERTIES OUTPUT_LOCATION "${CMAKE_CURRENT_SOURCE_DIR}/res/lang")
-qt5_add_translation(QM_FILES ${TS_FILES} OPTIONS -silent)
+if (Qt5_FOUND)
+    qt5_add_translation(QM_FILES ${TS_FILES} OPTIONS -silent)
+else()
+    qt_add_translation(QM_FILES ${TS_FILES} OPTIONS -silent)
+endif()
 
 # The following file includes X11 headers that later break Qt headers.
 set_source_files_properties("src/DisplayIsValid.cpp" PROPERTIES SKIP_UNITY_BUILD_INCLUSION TRUE)
@@ -187,7 +199,21 @@ add_executable (input-leap WIN32
 
 include_directories (./src)
 
-target_link_libraries(input-leap net arch base io Qt5::Core Qt5::Widgets Qt5::Network ${OPENSSL_LIBS})
+target_link_libraries(input-leap
+    net
+    arch
+    base
+    io
+    Qt${QT_VERSION_MAJOR}::Core
+    Qt${QT_VERSION_MAJOR}::Widgets
+    Qt${QT_VERSION_MAJOR}::Network
+)
+
+if (Qt6_FOUND)
+    target_link_libraries(input-leap Qt6::Core5Compat)
+endif()
+target_link_libraries(input-leap ${OPENSSL_LIBS})
+
 target_compile_definitions (input-leap PRIVATE -DINPUTLEAP_VERSION_STAGE="${INPUTLEAP_VERSION_STAGE}")
 target_compile_definitions (input-leap PRIVATE -DINPUTLEAP_REVISION="${INPUTLEAP_REVISION}")
 
@@ -233,5 +259,5 @@ if(INPUTLEAP_BUILD_TESTS)
              WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
     target_include_directories(guiunittests PUBLIC ../../ext)
-    target_link_libraries(guiunittests gtest gmock Qt5::Core Qt5::Widgets Qt5::Network ${libs})
+    target_link_libraries(guiunittests gtest gmock Qt${QT_VERSION_MAJOR}::Core Qt${QT_VERSION_MAJOR}::Widgets Qt${QT_VERSION_MAJOR}::Network ${libs})
 endif()
--- a/src/gui/src/KeySequence.cpp
+++ b/src/gui/src/KeySequence.cpp
@@ -191,7 +191,11 @@ QString KeySequence::keyToString(int key)
         {
             case Qt::LeftButton: return "1";
             case Qt::RightButton: return "2";
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+            case Qt::MiddleButton: return "3";
+#else
             case Qt::MidButton: return "3";
+#endif
         default:
             break;
         }
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -44,7 +44,7 @@
 #include <QMessageBox>
 #include <QFileDialog>
 #include <QDesktopServices>
-#include <QDesktopWidget>
+#include <QRegularExpression>
 
 #if defined(Q_OS_MAC)
 #include <ApplicationServices/ApplicationServices.h>
@@ -367,7 +367,11 @@ void MainWindow::logOutput()
     if (cmd_app_process_)
     {
         QString text(cmd_app_process_->readAllStandardOutput());
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+        for (QString line : text.split(QRegularExpression("\r|\n|\r\n"))) {
+#else
         for (QString line : text.split(QRegExp("\r|\n|\r\n"))) {
+#endif
             if (!line.isEmpty())
             {
                 appendLogRaw(line);
@@ -404,7 +408,12 @@ void MainWindow::appendLogError(const QString& text)
 
 void MainWindow::appendLogRaw(const QString& text)
 {
-    for (QString line : text.split(QRegExp("\r|\n|\r\n"))) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    auto lines = text.split(QRegularExpression("\r|\n|\r\n"));
+#else
+    auto lines = text.split(QRegExp("\r|\n|\r\n"));
+#endif
+    for (const auto& line : lines) {
         if (!line.isEmpty()) {
             m_pLogWindow->appendRaw(line);
             updateFromLogLine(line);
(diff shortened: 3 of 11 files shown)


shymega:
I've checked out this PR to have a look into it. This may have to wait for a new major release though. Can you add a newsfragment, please? Thank you.

shymega:
@jpwhiting I've fixed a build error. You may wish to pull from your branch (I pushed there) - you'll need the `Qt6-Core5Compat` package now, which is now a dep.

shymega:
Rebased against `master`. Also, we need to think about how to approach this. Can we use Qt5 and Qt6 at the same time? Ubuntu 18.04, which the GH Actions targets (idk about Azure Pipelines), would need Qt 6 backported.

shymega:
@jpwhiting Could you rebase against mainline, and adjust merge conflicts, making sure to use the new classes? We've had a *major* refactor since this PR.

Also, what would be really cool, would be a `ifdef` guard to use either Qt 5 or Qt 6. This could be configured via CMake. I'm cautious that using Qt 6 in future releases _only_ will result in a lot of users being left behind.

jpwhiting:
Yeah, will do. Most likely this weekend.

jpwhiting:
Ok, I tried using the generic Qt::Core, etc. but that needs Qt 5.15 apparently. while Ubuntu 20.04, etc. have 5.12, will use the uglier Qt{QT_MAJOR_VERSION}::Core instead and see if that works.

src/gui/CMakeLists.txt
```diff
@@ -235,5 +246,5 @@ if(INPUTLEAP_BUILD_TESTS)
              WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 
     target_include_directories(guiunittests PUBLIC ../../ext)
-    target_link_libraries(guiunittests gtest gmock Qt5::Core Qt5::Widgets Qt5::Network ${libs})
+    target_link_libraries(guiunittests gtest gmock Qt${QT_MAJOR_VERSION}::Core Qt${QT_MAJOR_VERSION}::Widgets Qt${QT_MAJOR_VERSION}::Network ${libs})
```
p12tic, line 249: Could be abstracted by QT_LIBS variable which we could populate near find_package where we select Qt version.

src/gui/src/AppConfig.cpp
```diff
@@ -243,3 +245,44 @@ bool AppConfig::getAutoStart() { return m_AutoStart; }
 void AppConfig::setMinimizeToTray(bool b) { m_MinimizeToTray = b; }
 
 bool AppConfig::getMinimizeToTray() { return m_MinimizeToTray; }
+
+void AppConfig::setStartWithComputer(bool b)
```
p12tic, line 249: This needs to go in via separate PR.

src/gui/src/KeySequence.cpp
```diff
@@ -191,7 +191,7 @@ QString KeySequence::keyToString(int key)
         {
             case Qt::LeftButton: return "1";
             case Qt::RightButton: return "2";
-            case Qt::MidButton: return "3";
+            case Qt::MiddleButton: return "3";
```
p12tic, line 194: Which is the minimum version of Qt that has this API?

src/gui/src/AppConfig.cpp
jpwhiting: Yep, I'll split it out, good call. One moment.

src/gui/src/KeySequence.cpp
jpwhiting: https://doc.qt.io/qt-5/qt.html#MouseButton-enum Qt 5.15 seems to have it. I'm not sure about older versions...

src/gui/CMakeLists.txt
```diff
@@ -189,7 +193,14 @@ add_executable (input-leap WIN32
 
 include_directories (./src)
 
-target_link_libraries(input-leap net arch base io Qt5::Core Qt5::Widgets Qt5::Network ${OPENSSL_LIBS})
+if (NOT Qt6_FOUND)
```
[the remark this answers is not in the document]
jpwhiting: https://cmake.org/cmake/help/latest/command/option.html OPTION lets us create a boolean flag on/off only. What would you suggest the option be? PREFER_QT_6 with default to OFF or something? I think it's pretty straightforward to try to link against Qt6 and if that's not there try Qt5 instead. In what situation would an OPTION be preferrable?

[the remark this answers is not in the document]
shymega: I think @jpwhiting's proposed approach of trying Qt 6 first, then Qt 5 is sound. Let's not confuse users or developers. Making it configurable (in terms of @p12tic's suggestion) would also lead to potential user error, which means more bug reports and time wasted.

jpwhiting:
Looks like Ubuntu 18.04 (which was EOL last month...) is the only failure left. Or am I missing something?

shymega:
> Looks like Ubuntu 18.04 (which was EOL last month...) is the only failure left. Or am I missing something?

This seems to be the problem:  https://github.com/input-leap/input-leap/actions/runs/5048828318/jobs/9060959082?pr=1469#step:10:627

I know we have a CI revamp PR pending... and the Azure CI checks can be ignored, they're not up to date. But they all seem to pass.

jpwhiting:
> > Looks like Ubuntu 18.04 (which was EOL last month...) is the only failure left. Or am I missing something?
> 
> This seems to be the problem: https://github.com/input-leap/input-leap/actions/runs/5048828318/jobs/9060959082?pr=1469#step:10:627
> 
> I know we have a CI revamp PR pending... and the Azure CI checks can be ignored, they're not up to date. But they all seem to pass.

Yeah, that method is only in Qt 5.15 nothing older. so Ubuntu 18.04 doesn't have it.

shymega:
@jpwhiting Sorry for my late reply. I took a step back from Input Leap whilst interviews were ongoing. Is there an alternative to that method in pre-5.15 Qt?

jpwhiting:
> @jpwhiting Sorry for my late reply. I took a step back from Input Leap whilst interviews were ongoing. Is there an alternative to that method in pre-5.15 Qt?

No idea. sorry. Probably not. but the real question is does Input Leap need to support platforms that are EOL for some reason? If so, what's the reason?

shymega:
> > @jpwhiting Sorry for my late reply. I took a step back from Input Leap whilst interviews were ongoing. Is there an alternative to that method in pre-5.15 Qt?
> 
> No idea. sorry. Probably not. but the real question is does Input Leap need to support platforms that are EOL for some reason? If so, what's the reason?

I think the issue is that we need to balance the 'reach' of Input Leap, distributions, and security.

Personally, I'd rather drop support for older platforms that are EOL, I agree, but then there might be backlash for those on older platforms. Windows XP, Vista, 7, older macOS versions, etc. So it's a tricky balance to maintain.

I think we don't really need to support Ubuntu 18.04, to be honest. I would rather use modern coding practices, and embrace new technology.

@p12tic Can we drop Ubuntu 18.04? I can update @Cimbali on the CI issue if we decide to drop Ubuntu 18.04.

Cimbali:
> Is there an alternative to that method in pre-5.15 Qt?

From the docs:

> Returns the expression wrapped between the \A and \z anchors to be used for exact matching.

So I suppose `QRegularExpression("^...$")` is the same as `QRegularExpression(anchoredPattern("..."))`. Not tested though. (To be fair starting with `^.*` seems a bit weird?)

shymega:
To follow-up on this: I need to get in touch with @p12tic, but it's possible we could drop support for Ubuntu 18.04, resolving the blocker.

src/gui/src/MainWindow.cpp
```diff
@@ -443,19 +447,20 @@ void MainWindow::checkConnected(const QString& line)
 
 void MainWindow::checkFingerprint(const QString& line)
 {
-    QRegExp fingerprintRegex(".*peer fingerprint \\(SHA1\\): ([A-F0-9:]+) \\(SHA256\\): ([A-F0-9:]+)");
-    if (!fingerprintRegex.exactMatch(line)) {
+    QRegularExpression fingerprintRegex(QRegularExpression::anchoredPattern(".*peer fingerprint \\(SHA1\\): ([A-F0-9:]+) \\(SHA256\\): ([A-F0-9:]+)"));
```
Cimbali, line 450: Finally found how to write suggestions! Here’s my previous comment and the contents of  #1667 in an easier form:

```suggestion
    QRegularExpression fingerprintRegex("peer fingerprint \\(SHA1\\): ([A-F0-9:]+) \\(SHA256\\): ([A-F0-9:]+)$");
```

Essentially avoiding the function that just wraps a pattern in `^...$` by adding the useful end-of-line marker ourselves.

shymega:
@jpwhiting I've rebased this against `master` again. Looking at @Cimbali's code suggestion, I'd say that's a reasonable approach. Does the newer Qt API offer any benefits over it, though? Best to keep things simple...

jpwhiting:
> @jpwhiting I've rebased this against `master` again. Looking at @Cimbali's code suggestion, I'd say that's a reasonable approach. Does the newer Qt API offer any benefits over it, though? Best to keep things simple...

Your approach sounds fine. I didn't spend much thought on that at all apart from porting it to Qt6, if you want to keep Qt5 support that sounds reasonable to me.

shymega:
> > @jpwhiting I've rebased this against `master` again. Looking at @Cimbali's code suggestion, I'd say that's a reasonable approach. Does the newer Qt API offer any benefits over it, though? Best to keep things simple...
> 
> Your approach sounds fine. I didn't spend much thought on that at all apart from porting it to Qt6, if you want to keep Qt5 support that sounds reasonable to me.

OK, I'll commit that and push to the PR. Once we get @p12tic's approval, we can get this in for v3.0.0!

shymega:
I can change the wording of that commit, it's a bit late here in the UK, so my eyes are slowly drooping.

theofficialgman:
> Personally, I'd rather drop support for older platforms that are EOL, I agree, but then there might be backlash for those on older platforms. Windows XP, Vista, 7, older macOS versions, etc. So it's a tricky balance to maintain.
> 
> I think we don't really need to support Ubuntu 18.04, to be honest. I would rather use modern coding practices, and embrace new technology.
> 
> @p12tic Can we drop Ubuntu 18.04? I can update @Cimbali on the CI issue if we decide to drop Ubuntu 18.04.

I'd personally love is Ubuntu 18.04 could still be targeted. I have to use it for multiple reasons on Nvidia Tegra ARM64 hardware and I know there is a lot of that in use out there. I personally use barrier all the time when developing on this hardware.

shymega:
I can see the use-case, but on the other hand - it is legacy, and we can't be expected to support 18.04 for much longer.

I think the community should signal which way we should go - legacy or modern.

theofficialgman:
Well you are welcome to use the multiple backport ppas that exist. I don't know what the intended linux packaging targets are for release.

depending on the QT licences and compatibility with input-leap any of these bionic ppas are options:
QT 5.15.2 (i386 and amd64): https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.15.2-bionic
QT 5.15.2 (armhf and arm64): https://launchpad.net/~theofficialgman/+archive/ubuntu/opt-qt-5.15.2-bionic-arm/+packages
QT 6.2.2 (i386, amd64, armhf, arm64): https://launchpad.net/~okirby/+archive/ubuntu/qt6-backports/+packages?field.name_filter=&field.status_filter=published&field.series_filter=bionic

the QT6 ppa in particular I know does not have all the plugins backported, but it probably has everything input-leap would ever use.

if you intend on providing a package for ubuntu bionic as a deb of appimage, you would have to bundle QT into it.

shymega:
> Well you are welcome to use the multiple backport ppas that exist. I don't know what the intended linux packaging targets are for release.
> 
> depending on the QT licences and compatibility with input-leap any of these bionic ppas are options: QT 5.15.2 (i386 and amd64): https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.15.2-bionic QT 5.15.2 (armhf and arm64): https://launchpad.net/~theofficialgman/+archive/ubuntu/opt-qt-5.15.2-bionic-arm/+packages QT 6.2.2 (i386, amd64, armhf, arm64): https://launchpad.net/~okirby/+archive/ubuntu/qt6-backports/+packages?field.name_filter=&field.status_filter=published&field.series_filter=bionic
> 
> the QT6 ppa in particular I know does not have all the plugins backported, but it probably has everything input-leap would ever use.

We will look into this.

> 
> if you intend on providing a package for ubuntu bionic as a deb or appimage, you would have to bundle QT into it.

AppImage is unlikely to happen.
 
> I personally feel that input-leap has simple job and should stay that way and should support the widest range of platforms possible.

This is incorrect. It's not simple. We also intend to support Wayland, and we have to drop support for platforms _at some point_. It's unavoidable. It's a complex codebase, and whilst we're looking into porting to Rust, this is a long way off.

theofficialgman:
> This is incorrect. It's not simple. We also intend to support Wayland, and we have to drop support for platforms _at some point_. It's unavoidable. It's a complex codebase, and whilst we're looking into porting to Rust, this is a long way off.

simple as in a simple job (like the unix philosophy, do one thing and do it well). not simple as in a simple codebase. the job is sharing keyboard/mouse and clipboard between multiple devices.

shymega:
> > This is incorrect. It's not simple. We also intend to support Wayland, and we have to drop support for platforms _at some point_. It's unavoidable. It's a complex codebase, and whilst we're looking into porting to Rust, this is a long way off.
> 
> **simple job** (like the unix philosophy, do one thing and do it well). not simple as in a simple codebase. the job is sharing keyboard/mouse and clipboard between multiple devices. job simplicity is even in your own description:
> 
> > Input Leap aims to maintain that simplicity. Input Leap will let you use your keyboard and mouse from one computer to control one or more other computers.
> > Clipboard sharing is supported.
> > That's it.

Ah, I understand now. I think we need to update our MO in the README - sorry for my obtuseness, I can read things literally.. a lot, it's my downfall.

shymega:
For now, I think dropping Ubuntu 18.04 would be wise. It's EOL now, and as Canonical declares EOL, smaller projects will not be expected to retain support for it.

I want to get this merged soon...

shymega:
Rebased against mainline again.

p12tic:
Finally, we can have this PR in too.

shymega:
> Finally, we can have this PR in too.

The only couple of things I would like to add is, to make sure users can still use Qt 5 - did we test that? And also, the commit in #1695, should have been in this PR, as it's part of the scope for Qt 6. 

I think we should also make this a breaking change - but as we haven't released v3.0.0 yet, I'm not sure it matters too much.

PR #1478, opened 2022-06-29
Generate an appdata file and install it

whot:
We've generated the appdata file in the (reference?) spec file but we might as well just generate this as part of our build and install it. And while we're there, let's install our man pages too.

## Contributor Checklist:

  - *Note by @whot: Don't know if the above counts as user-visible change?*


p12tic (approved):

p12tic:
CI failure seems to be related to github infrastructure, nothing we can fix.

PR #1479, opened 2022-06-29
Fix the reference rpm spec file

whot:
Sits on top of #1478

Fix the reference spec file we have in `dist/rpm` to actually work, and to work from the source directory. We can now run

```
        mkdir build
        cd build
        cmake ..
        make package_source
        rpmbuild -D="%_sourcedir $PWD" -bb rpm/*.spec
```
and that will (on F36 at least) build our package as rpm. A new CI job was added to run this as part of the CI.

That job probably won't work yet but I can't test it without permissions :)

## Contributor Checklist:


p12tic:
I rebased PR on top of later master and fixed an issue with github not being able to download git submodules.

whot:
ok, got a bit carried away here but this works now :)

The new job builds the rpm based on the source package generated by cmake, then builds both a binary and source rpm and stores those as artefacts. The binary one is obviously Fedora 36 (or whatever is used in the workflow) but the source package could be rebuild anywhere locally with `rpmbuild --rebuild $package.src.rpm`.

The rpm revision is derived from date and time, so the most recently built package will sort highest. This makes it immediately possible to tell a (Fedora or rpm-based distro) user to "grab that rpm from $PR and dnf install it for testing".

p12tic (approved):

PR #1487, opened 2022-08-04
A few whitespace fixes

whot:
Some really old commits I had in my barrier branch, copied over to IL now. The files modified for whitespace are just some random ones I looked at. I don't disagree with the new approach without alignment, but IMO this should be done all at once rather than mixing it.

Verify that only whitespace changed with `git diff --ignore-all-space HEAD~`.


whot:
CI failure is unrelated, in the artifact upload bits
> Error: Unexpected response. Unable to upload chunk to https://pipelines.actions.githubusercontent.com/otTl29S8mHkJ5BHuDxTOqZ0ujOmKAUj4vsHnfuHjVqiQZ5fLq8/_apis/resources/Containers/2455840?itemPath=Windows+Debug%5CDebug.zip

PR #1488, opened 2022-08-04
A little bit of CMake cleanup

whot:
Not sure this counts as "user-visible change" since most users definitely won't see this and most builders either :smile: 

This cleanup vastly simplifies the X11 package checks by using pkgconfig instead. I think at this point we can expect this to exist and work on every system from the last decade. automake and meson are exclusively requiring it, so not using it just makes our life harder.

Plus, most of the checks here were borderline pointless a decade ago, they certainly are now. There are no X servers anymore that don't support XKB/XI2/Xrandr for example, and virtually all desktop systems have all libX* installed anyway.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -41,7 +41,7 @@ jobs:
           add-apt-repository ppa:git-core/ppa
           apt-get update -y
           apt-get dist-upgrade -y
-          apt-get install -y git g++ libxtst-dev qtdeclarative5-dev libavahi-compat-libdnssd-dev libcurl4-openssl-dev libssl-dev cmake ninja-build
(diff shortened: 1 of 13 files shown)


whot:
ok, a bit of CI trouble but it all works now.

I'm still somewhat of a CMake beginner, so if for some reason `find_library()` is superior please let me know.

I picked `FOO_LINK_LIBRARIES` since it uses the full path to the .so file so it works non-default library paths (I use `/opt` a lot). In theory this means we can get rid of the BSD bits too but I don't want to touch that... Without that, we'd have to add `target_link_directories` everywhere.

Unfortunately though `LINK_LIBRARIES` is missing in Ubuntu 18.04. That was fixed shortly after that release but it missed the boat. So there's a check for that with a fall back to normal `LIBRARIES`, however that does mean if you have custom PKG_CONFIG_PATH or library paths they won't work well on 18.04. Then again, not sure this used to work before so there's that :)

p12tic (approved):

p12tic:
> if you have custom PKG_CONFIG_PATH or library paths they won't work well on 18.04.

I think we can ignore this problem for now.

PR #1491, opened 2022-08-08
cmake: big ifdef cleanup

whot:
A lot of these checks come from 20y ago when the unix landscape was quite different. Now things tend to be a lot more standardised and we can assume the existence of various header files, type sizes, etc.


whot:
ci-failure is unrelated, 503 during artifact upload

p12tic (approved):

PR #1494, opened 2022-08-08
Fix the hanging integrity tests

whot:
Also makes `barrierc` respond to SIGINT again which is a rather nice (developer) UX :)

Fixes #1490

cc @ofourdan

Issue #1490, opened 2022-08-05
Switch to standard mutexes has integtests hanging


whot:
### What happened?

One of the commits to switch to standard mutexes causes `integtests` to hang. Last good commit is ce894dca62069e6b1c9cd14e7e1e531b48fbcb80, first bad commit is 9fb908b1a22c2f7b9a067b2f7d25dc616f29ea0b but the three commits in between aren't immediately testable because of a compiler error.

*edit:* Narrowed down to 94635df43c95f64d8d9983bbd3ca58544cd77c21 as the bad commit after cherry-picking 9fb908b1a22c2f7b9a067b2f7d25dc616f29ea0b during the bisect to get past the compiler errors.

Reproduce with: `./build/bin/integtests`, go have a coffee, come back and notice the coffee cup is empty but the tests are still hanging.

`git bisect log` is below.

### Version

From Git HEAD or commit (specify below)

### Git commit hash (if applicable)

708cbfded015e641a3ffd43927f37e9879d3c1c4

### What OSes are you seeing the problem on? (Check all that apply)

Linux

### What OS versions are you using?

Fedora 36

### Relevant log output

```shell
$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# bad: [708cbfded015e641a3ffd43927f37e9879d3c1c4] Merge pull request #1487 from whot/wip/whitespace-fixes
git bisect bad 708cbfded015e641a3ffd43927f37e9879d3c1c4
# status: waiting for good commit(s), bad commit known
# good: [b0f0a6fa96af56ae6b0e9d1951efc0039178b8ac] Merge pull request #714 from p12tic/use-std-string-6
git bisect good b0f0a6fa96af56ae6b0e9d1951efc0039178b8ac
# good: [a9b30951ce6762353e28e9068af5364778f1fcb2] lib: Add utility function to convert from hex to binary
git bisect good a9b30951ce6762353e28e9068af5364778f1fcb2
# good: [0fbfed22a8c5c7a1e2b3318c7ef6c4fd114cd260] Rename version-related config macros from BARRIER* to INPUTLEAP*
git bisect good 0fbfed22a8c5c7a1e2b3318c7ef6c4fd114cd260
# bad: [c35c4bf148cad1f31b00d60123e8e7e35939b600] Merge pull request #1447 from mokibit/misc-warnings
git bisect bad c35c4bf148cad1f31b00d60123e8e7e35939b600
[...]

PR #1495, opened 2022-08-08
Hook up the tests via cmake and in the github workflows

whot:
github workflow is for the linux builds only atm. And obviously they fail because of #1490, so waiting with this PR until #1494 is merged is best.

### DIFF
--- a/clean_build.sh
+++ b/clean_build.sh
@@ -4,26 +4,28 @@ cd "$(dirname "$0")" || exit 1
 
 # some environments have cmake v2 as 'cmake' and v3 as 'cmake3'
 # check for cmake3 first then fallback to just cmake
-
-B_CMAKE=$(command -v cmake3 2>/dev/null)
-if [ "$?" -eq 0 ]; then
-    continue
-else
-    if command -v cmake 2>/dev/null; then
-        B_CMAKE=$(command -v cmake)
-    else
-        echo "ERROR: CMake not in $PATH, cannot build! Please install CMake, or if this persists, file a bug report."
-        exit 1
-    fi
+[ -n "$B_CMAKE" ] || B_CMAKE=$(command -v cmake3)
+[ -n "$B_CMAKE" ] || B_CMAKE=$(command -v cmake)
+if [ -z "$B_CMAKE" ]; then
+    echo "ERROR: CMake not in $PATH, cannot build! Please install CMake, or if this persists, file a bug report."
+    exit 1
 fi
 
-CMAKE_VERSION="$("$B_CMAKE" --version)"
-if [ "$(printf '%s\n' "3.12.0" "$currentver" | sort -V | head -n1)" = "3.12.0" ]; then
-    CMAKE_HAS_PARALLEL=1
+if [ "$(uname)" = "Darwin" ]; then
+    nproc=$(sysctl -n hw.ncpu)
 else
-    CMAKE_HAS_PARALLEL=0
+    nproc=$(nproc)
+fi
+if [ -z "$nproc" ]; then
+    nproc=8
+fi
+if [ $nproc -gt 32 ]; then
+    nproc=32
 fi
 
+export CMAKE_BUILD_PARALLEL_LEVEL=${nproc}
+
+B_BUILD_DIR="${B_BUILD_DIR:-build}"
 B_BUILD_TYPE="${B_BUILD_TYPE:-Debug}"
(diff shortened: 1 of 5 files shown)


clean_build.sh
```diff
@@ -4,26 +4,23 @@ cd "$(dirname "$0")" || exit 1
...
+    exit 1
 fi
 
-CMAKE_VERSION="$("$B_CMAKE" --version)"
-if [ "$(printf '%s\n' "3.12.0" "$currentver" | sort -V | head -n1)" = "3.12.0" ]; then
-    CMAKE_HAS_PARALLEL=1
-else
-    CMAKE_HAS_PARALLEL=0
+# Let's be reasonably nice in shared envs like CI runners
+nproc=$(nproc)
```
p12tic, line 16: This script is also used on OSX, `nproc` is not available there.

clean_build.sh
```diff
@@ -4,26 +4,23 @@ cd "$(dirname "$0")" || exit 1
...
-    continue
-else
-    if command -v cmake 2>/dev/null; then
-        B_CMAKE=$(command -v cmake)
-    else
-        echo "ERROR: CMake not in $PATH, cannot build! Please install CMake, or if this persists, file a bug report."
-        exit 1
-    fi
+[ -n "$B_CMAKE" ] || B_CMAKE=$(command -v cmake3)
+[ -n "$B_CMAKE" ] || B_CMAKE=$(command -v cmake2)
```
p12tic, line 8: `cmake2` will probably refer to cmake v2.x if present. This won't work, because the project has `cmake_minimum_required (VERSION 3.4). I agree with the change of checking style otherwise.

clean_build.sh
```diff
@@ -41,23 +38,15 @@ fi
...
 # Initialise Git submodules
 git submodule update --init --recursive
 
-rm -rf build
-mkdir build || exit 1
-cd build || exit 1
-echo "Starting Input Leap $B_BUILD_TYPE build..."
-"$B_CMAKE" $B_CMAKE_FLAGS .. || exit 1
-
-if [ "$CMAKE_HAS_PARALLEL" -eq "1" ]; then
```
p12tic, line 53: This complication here is to support building on older distributions. E.g. Ubuntu 18.04 only has cmake 3.10, whereas `--parallel` and friends became supported since 3.12.

whot: yep, but [the documentation](https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-tool-mode) points at `CMAKE_BUILD_PARALLEL_LEVEL` being the override when `--parallel` is not given:

> The [CMAKE_BUILD_PARALLEL_LEVEL](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html#envvar:CMAKE_BUILD_PARALLEL_LEVEL) environment variable, if set, specifies a default parallel level when this option is not given.

iow, we no longer need the check for versions because old cmake won't care about some random env var and new cmake handles it correctly.

clean_build.sh
whot: Fixed!

clean_build.sh
whot: fixed by running `sysctl` on darwin

whot:
Fixed the comments above (and the commit message which still referred to changes I had to drop since). I dropped the commit to add the test suite to the github workflow though because right now it fails and let's not have a broken pipeline until we can figure out what the test case failure is. I filed #1497 for that commit now.

clean_build.sh
p12tic: Indeed, your solution is equivalent to the previous code in less lines. I missed that `ninja` would run using all cores anyway.

p12tic (approved):

clean_build.sh
whot: > I missed that ninja would run using all cores anyway.

so, interestingly enough: we're now actually *limiting* the number of cores to be used. I picked 32 as arbitrary max to be nice on beefy runners (on gitlab.fdo we have `FDO_CI_CONCURRENT` that specifies how many you can use in a job).

If we just want to run with whatever is the default, we can drop this code altogether because then ninja will just use whatever its built-in default is. We could export `MAKEFLAGS=-j` to ensure make does run in parallel where we fall back to that. but for ninja, no work is needed.

p12tic: That's good point, we don't need unnecessary limiting. Myself I thought the reason you wanted that is to have a limit on some CI. Given that 32 threads is plenty for such a small project, I didn't think much about this limitation. But if there's no need for it after all, let's remove it.

whot: Done in #1502

PR #1497, opened 2022-08-08
Run the tests during the github workflow

whot:
This is on top of #1495, just the last commit to add it to the github CI.


shymega:
I'd rather you push just the commit for running tests to this PR, otherwise we'll have a conflict. Alternatively, you could just push the running tests in CI commit to the original PR, and I can close this one.

shymega:
> @shymega This is not a problem, we just need to rebase on latest master and the PR will contain only one remaining commit.

Oh. It doesn't make a lot of sense to make two PRs with the same lot of commits, but alright - good plan there.

whot:
> It doesn't make a lot of sense to make two PRs with the same lot of commits

Mostly depends on the situation. Usually when I do this is because there's some dependency on the earlier commits, but the pieces are logically separate. Like in this case, it doesn't make sense to merge this PR unless you want to stare at a broken pipeline for every future PR but it made sense to merge #1495.

[And an earlier version of this commit had a hard dependency on the earlier commit, I didn't click the final version was completely standalone.]

Where projects rebase instead of merge-commit the identical commits just disappear. In the merge-commit case there shouldn't be conflicts, git is smart enough to know they're the same commits.


Aaaanyway - I've rebased this now, so this commit is ready to... not go in because the pipeline fails :)

whot:
Well, that was... "fun". But the tests build and run in the CI now (at least in Linux), should be ready to go in

p12tic:
Cool, thanks a lot!

p12tic (approved):

PR #1498, opened 2022-08-09
treewide: drop the whitespace alignment for class member variables

whot:
Realign all header files to use a single space between the class member
type and its name, i.e. "int foo;" instead of the previous attempted
alignment of all member variables to the same tabstop.

Patch created by running:
```
 sed -i 's/ \+\(m_.*\);/ \1;/' src/**/*.h
```
and then fixing/ignoring the occasional outlier.

This is in response to https://github.com/input-leap/input-leap/pull/1494#issuecomment-1208113958. Best to do this in one go all over the tree to have some sort of consistency (even though this patch probably doesn't catch all occurrences).

## Verification

Only thing `git diff -w` picks up is a manually added line join

```
$ git diff -w HEAD~
diff --git a/src/lib/platform/MSWindowsScreen.h b/src/lib/platform/MSWindowsScreen.h
index be9bbd3d..63864c8a 100644
--- a/src/lib/platform/MSWindowsScreen.h
+++ b/src/lib/platform/MSWindowsScreen.h
@@ -282,8 +282,7 @@ private:
     HKL m_keyLayout;
 
     // screen saver stuff
-    MSWindowsScreenSaver*
-                        m_screensaver;
+    MSWindowsScreenSaver* m_screensaver;
     bool m_screensaverNotify;
     bool m_screensaverActive;
 
@@ -333,8 +332,7 @@ private:
 
     mutable std::string m_dropTargetPath;
 
-    MSWindowsDropTarget*
-                        m_dropTarget;
+    MSWindowsDropTarget* m_dropTarget;
     HWND m_dropWindow;
     const int m_dropWindowSize;
 


```

### DIFF
--- a/src/cmd/barrierc/MSWindowsClientTaskBarReceiver.h
+++ b/src/cmd/barrierc/MSWindowsClientTaskBarReceiver.h
@@ -56,11 +56,11 @@ private:
                             UINT msg, WPARAM wParam, LPARAM lParam);
 
 private:
-    HINSTANCE            m_appInstance;
-    HWND                m_window;
-    HMENU                m_menu;
-    HICON                m_icon[kMaxState];
-    const BufferedLogOutputter*    m_logBuffer;
+    HINSTANCE m_appInstance;
+    HWND m_window;
+    HMENU m_menu;
+    HICON m_icon[kMaxState];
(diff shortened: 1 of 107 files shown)


p12tic (approved):
Thanks, this is much fewer changes than I expected.

PR #1502, opened 2022-08-11
Drop forced parallel job count in clean_build.sh

whot:
Let's rely on cmake just using the native tool's job count


whot:
Note: I'm installing ninja now too on macos. This means we don't really have a CI test for Makefiles anymore but... meh? cmake is supposed to generate these for us, so any bug there would be a cmake bug anyway.

p12tic (approved):

PR #1504, opened 2022-08-16
Drop a leftover mention of the xinitthreads

whot:
Missing from 7b85e30c, must have gotten lost during a rebase.

Original PR: #1503


shymega (approved):
This is fine, no objections. Merging now.

PR #1505, opened 2022-08-16
Rework the argument parser

whot:
Since we can't use simple `getopt` because of windows, let's at least make the handling a lot less convoluted using a wrapper class around what is just a `std::queue`. 

All the args get pushed into the queue, then we can either `shift()` them along or conditionally `shift("--foo", "-bar", &optarg)` them. This makes the code less nested and a lot easier to read (and extend).


p12tic (approved):
Thanks!

PR #1522, opened 2022-11-26
Make dist workflow only build on Input Leap tag

shymega:
This modification reduces confusion for users with GH Actions, and I will be looking to make generic amd64/arm64 artifacts for debug builds on Linux/Mac, but for now, this fix makes it so the `dist.yml` workflow only builds Fedora RPMs on a tag release. That way, we have only one CI workflow actually running when it comes to new commits, and publishing generic amd64/arm64 Mac, Linux, and amd64 Windows executable artifacts.

Thoughts?

## Contributor Checklist:

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   fedora:
+    if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
     container: fedora:36
 


p12tic:
But it still makes sense to build on fedora as part of the CI to see failures in case tests don't work with the particular set of dependency versions available on that distribution. What is the specific confusion of GH actions users that you're referring to?

shymega:
@p12tic See [here](https://github.com/input-leap/input-leap/discussions/1517#discussioncomment-4194507) - we currently have a distribution-specific builder, that builds for Fedora, but it's not under the CI workflow. To me, when I look at Actions, I see two separate builds, and it should be combined into one workflow. That way we can link directly on the README/wiki, and then it's clear which one has Debug artifacts or not. We don't really need to produce RPMs for debug builds, as the general purpose - iirc - is to produce binaries.

shymega:
Just to add to my last sentence, the other thing is that we don't *really* need to build RPMs or DEBs ourselves. I intend to take over the Input Leap AUR package as it's now orphaned, but we should leave packaging to distribution maintainers, and provide the tooling for them to do so.

whot:
>  we should leave packaging to distribution maintainers,

fully agree with that statement. 

the only reason this is hooked into the CI is because the spec file was already there, so building it in the CI at least makes sure that part doesn't break.

> we have only one CI workflow actually running when it comes to new commits

that bit is IMO a misunderstanding of how github workflows work. there's no real need to reduce it to one workflow other than "that's what I expect". workflows are more about technical logical grouping than anything else. 

To avoid the confusion in the other issue, renaming the workflow to "distribution checks" would be 90% of the way there. Likewise, renaming it from `CI` to `build tests` gets rid of the "only one workflow for the CI" need - the CI in github are *all* the workflows together.

.github/workflows/dist.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 jobs:
   fedora:
+    if: startsWith(github.ref, 'refs/tags/')
```
whot, line 12: ftr, if you don't build this with every push or PR, you won't find out that it's broken until after the tag is pushed. And by then it's too late, because fixing the spec file would require you to re-tag and that's not really worth it.

whot:
I forgot - one thing this CI *could* be useful for is to effectively produce something akin to a nightly repo. But that would require maintainer time, so it's better punted to ... actual distro maintainers :)

shymega:
> > we should leave packaging to distribution maintainers,
> 
> fully agree with that statement.
> 
> the only reason this is hooked into the CI is because the spec file was already there, so building it in the CI at least makes sure that part doesn't break.

Perhaps we should make an issue to track progress on *removing* the RPM and DEB scaffolding then? If we provide stable code to distribution maintainers, and make sure they have the support they need (i.e, IRC/Matrix on Libera, possibly a mailing list...?) to supply packages to users, then surely we don't need to maintain the packaging scaffolds?

> 
> > we have only one CI workflow actually running when it comes to new commits
> 
> that bit is IMO a misunderstanding of how github workflows work. there's no real need to reduce it to one workflow other than "that's what I expect". workflows are more about technical logical grouping than anything else.

Yeah, I suppose I conflated the terminology together. Sorry about that. I'm used to dealing with Jenkins.

> To avoid the confusion in the other issue, renaming the workflow to "distribution checks" would be 90% of the way there. Likewise, renaming it from `CI` to `build tests` gets rid of the "only one workflow for the CI" need - the CI in github are _all_ the workflows together.

Yup.

shymega:
> I forgot - one thing this CI _could_ be useful for is to effectively produce something akin to a nightly repo. But that would require maintainer time, so it's better punted to ... actual distro maintainers :)

Yes, I don't particularly want to provide Linux nightlies, but for Mac and Windows, I think that's something we should look into. For Linux, it's made more difficult with `glibc` version mismatches, or other shared libraries. I know dependency hell is a thing on Windows as well, but I've never had as many issues than I have with Linux.

shymega:
Keeping this PR open for *now* until we finish up the discussion and merge #1525.

whot:
> For Linux, it's made more difficult with glibc version mismatches, or other shared libraries

On the plus side, you can flatpak input leap to avoid some/most of this. Doubly so with #1524 which will work under Wayland (for supported compositors). But anyway... #1525 is filed now (not it'll conflict once this one here is merged, I'll fix that then).

PR #1524, opened 2022-12-08
Draft: add support for libei

whot:
I'm filing this PR just so it's sitting somewhere public instead of on a private branch that only few people know.

**This branch is subject to rebases**, in fact I'll probably close this PR at some point and open a new one with a cleaned-up branch instead.

libei is a library for **E**mulated **I**nput and a replacement for what is XTEST under X11. Unlike XTEST it's windowing-system agnostic, it's just a transport layer for sending *logical* input events. See the [libei README](https://gitlab.freedesktop.org/libinput/libei/)  for details on the motivation etc.

libei clients like input-leap can work as receiver context (barriers, real input events are sent to us via libei) or as sender context (barrierc, emulated events are sent by us to the compositor). However, libei is only the transport layer, the logic of establishing a connection is punted to the portals, in particular the `RemoteDesktop` portal for barrierc and the new `InputCapture` portal for barriers.

For the portal communication we're using libportal which requires a GLib mainloop (and thus a separate thread).

This PR adds the required bits to:
- make X11 optional and add libei as a separate backend (`--use-ei` in barrierc/barriers)
- add a GLib mainloop in a separate thread so we can use libportal easily
- add the bits to `barrierc` to talk to [org.freedesktop.portal.RemoteDesktop.ConnectToEIS](https://github.com/flatpak/xdg-desktop-portal/pull/762) to set up libei and then use those devices to send input events
- add the bits to `barriers` to talk to [org.freedesktop.portal.InputCapture](https://github.com/flatpak/xdg-desktop-portal/pull/714) to set up pointer barriers and get input events from the compositor

To actually run this, you will need to replace a bunch of components, see the [tracking issue on libei](https://gitlab.freedesktop.org/libinput/libei/-/issues/1). As of opening the PR, you will need updated mutter, libportal, xdg-desktop-portal, xdg-desktop-portal-gnome and of course libei itself.

The input-leap internal bits are unchanged, so you can use a libei-enabled barriers to talk to input-leap's barrierc on Win/MacOs/X11 and so on.

cc @ofourdan who deserves most of the credit for the hard parts

### DIFF
--- a/src/lib/platform/EiKeyState.cpp
+++ b/src/lib/platform/EiKeyState.cpp
@@ -38,8 +38,6 @@ EiKeyState::EiKeyState(EiScreen* screen, IEventQueue* events) :
     // one during initial startup - even before we know what our actual keymap is.
     // Once we get the actual keymap from EIS, we swap it out so hopefully that's enough.
     init_default_keymap();
-
-    xkb_state_ = xkb_state_new(xkb_keymap_);
 }
 
 void EiKeyState::init_default_keymap()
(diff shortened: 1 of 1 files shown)


p12tic:
Replaced by https://github.com/input-leap/input-leap/pull/1594.

PR #1526, opened 2022-12-09
Remove the rpm and debian packaging files

whot:
Better to push responsibility for Input Leap builds to distribution maintainers which have better facilities for this anyway.

See https://github.com/input-leap/input-leap/pull/1522#issuecomment-1342937008

## Contributor Checklist:


truatpasteurdotfr:
I would leave the deb/rpm build files to give the distribution persons something to start with, and document that these are legacy files. If they are still building from CI, good leave them and make these build  failure non blocking. Just my 2 cents.

p12tic:
I think these are useful for testing locally, because a hacky deb or rpm is still better than `make && make install`. In case of Debian it makes sense to move the files e.g. to `dist` subfolder so that Debian people don't need to patch the distribution tarball.

p12tic:
I moved the debian files to dist/ in https://github.com/input-leap/input-leap/pull/1537. We currently don't use them, but it makes sense to build a deb and rpm files on every PR.

shymega:
> I moved the debian files to dist/ in #1537. We currently don't use them, but it makes sense to build a deb and rpm files on every PR.

Does it make sense, though? Is it our responsibility? Personally, I think there's so many distro-specific quirks and nuances, we shouldn't keep the DEB or RPM files. 

I'm focusing on getting Flatpak & Snap available soon, but I think that should be our limit TBH.

whot:
>  because a hacky deb or rpm is still better than `make && make install`

tbh, i'd say it'd be better to ensure that users can easily run from the build directory so there's no need to even `make install`. This way you get the best of all worlds - users can run the current one without having to wait for a distro package, they cannot mess up their system installation, and it's easy for users to bisect any bugs.

PR #1527, opened 2022-12-13
Update gui.ts for v2.4.0

maboroshin:
Add/Modify UI String for 2.4.0.

I am a Japanese translator. https://github.com/debauchee/barrier/pull/536

I will update the original English string ```gui.ts``` for translation for v2.4.0.

## Contributor Checklist:

### DIFF
--- a/src/gui/gui.ts
+++ b/src/gui/gui.ts
@@ -41,6 +41,11 @@ The Barrier GUI is based on QSynergy by Volker Lanz.
         <translation type="unfinished"></translation>
     </message>
     <message>
+        <location filename="res/AboutDialogBase.ui" line="167"/>
+        <source>Build Date: </source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
         <location filename="res/AboutDialogBase.ui" line="163"/>
         <source>&amp;Ok</source>
         <translation type="unfinished"></translation>
@@ -141,6 +146,11 @@ The Barrier GUI is based on QSynergy by Volker Lanz.
     <message>
         <location filename="res/ActionDialogBase.ui" line="264"/>
         <source>the hotkey is released</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="res/ActionDialogBase.ui" line="148"/>
+        <source>Toggle screen</source>
         <translation type="unfinished"></translation>
     </message>
 </context>
@@ -356,7 +366,7 @@ Do you want to install Bonjour?</source>
     </message>
     <message>
         <location filename="src/MainWindow.cpp" line="408"/>
-        <source>Barrier is now connected, You can close the config window. Barrier will remain connected in the background.</source>
+        <source>Barrier is now connected. You can close the config window and Barrier will remain connected in the background.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


src/gui/gui.ts
```diff
@@ -559,6 +574,26 @@ To automatically trust this fingerprint for future connections, click Yes. To re
         <source>Run Wizard</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <location filename="res/MainWindowBase.ui" line="435"/>
+        <source>S&amp;ave configuration</source>
```
shymega, line 579: Should there be an ampersand here?

src/gui/gui.ts
```diff
@@ -370,7 +380,7 @@ Do you want to install Bonjour?</source>
 
 %1
 
-This is a server fingerprint. You should compare this fingerprint to the one on your server&apos;s screen. If the two don&apos;t match exactly, then it&apos;s probably not the server you&apos;re expecting (it could be a malicious user).
+This is a server fingerprint. You should compare this fingerprint to the one on your server's screen. If the two don't match exactly, then it's probably not the server you're expecting (it could be a malicious user).
```
shymega, line 383: This is mostly fine, but I'd rather we kept the `&apos;` here, as it's elsewhere in translations. Continuity, so to speak.

shymega:
Overall, it looks fine to me, apart from the comments I've made in my review. I will wait for @p12tic to review too.

maboroshin:
The origin of these strings is the source code. The string in the ```.ts``` must match the string in the source code. Nothing happens for strings that exist only in ```.ts```. The English ```gui.ts``` is the prototype to be copied. All language ```.ts``` need to be updated.

The program does not refer to ```filename``` and ```line``` in the location tag. This is seen by humans. And line numbers change frequently, so approximate is fine.

I checked about review for the fingerprint's string. (Line 373 383) This string will no longer exist in the next version for the next version after 2.4.0. The following changes are required on ```gui.ts```. This allows the translator to notice updates to the English strings and translate each ```.ts```.

Delete below:
```
    <message>
        <location filename="src/MainWindow.cpp" line="434"/>
        <source>Security question</source>
        <translation type="unfinished"></translation>
    </message>
    <message>
        <location filename="src/MainWindow.cpp" line="435"/>
        <source>Do you trust this fingerprint?

%1

This is a server fingerprint. You should compare this fingerprint to the one on your server's screen. If the two don't match exactly, then it's probably not the server you're expecting (it could be a malicious user).

To automatically trust this fingerprint for future connections, click Yes. To reject this fingerprint and disconnect from the server, click No.</source>
        <translation type="unfinished"></translation>
    </message>
```

Add below:
```
    <message>
        <location filename="src/FingerprintAcceptDialog.ui" line="20"/>
        <source>Security question</source>
        <translation type="unfinished"></translation>
    </message>
    <message>
        <location filename="src/FingerprintAcceptDialog.ui" line="45"/>
        <source>SHA1 (deprecated, compare to old servers only)</source>
        <translation type="unfinished"></translation>
    </message>
    <message>
        <location filename="src/FingerprintAcceptDialog.ui" line="85"/>
        <source>Do you trust this fingerprint?</source>
        <translation type="unfinished"></translation>
    </message>
```

src/gui/gui.ts
p12tic: This comes from MainWindowBase.ui. Looks like ampersand has a special meaning in Qt.

PR #1531, opened 2022-12-20
enable upload macos build to artifacts

kukat:
## Contributor Checklist:

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -102,7 +102,6 @@ jobs:
           name: Publish the DMG for ${{ matrix.os }}
           path: ./build/bundle/*.dmg
           if-no-files-found: warn
-        if: ${{ env.B_BUILD_TYPE == 'Release' }}
 
   win-build-2019:
     runs-on: windows-2019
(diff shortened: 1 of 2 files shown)


shymega:
@kukat You need to document these commits more, they're very vague. Also, I've restored the PR template - please don't delete that. If you don't feel it needs ticking; just don't tick it, but don't delete the template.

I've made a review for now, and checks have passed.

shymega:
@p12tic Should we close this PR, or modify the commit message, fix merge conflicts, and then merge manually?

p12tic:
Looking second time into it I don't understand the purpose. If we want dmg to be uploaded, we should just do release build and that's it. The build difference between Debug and Release has a significant difference in that Release strips symbols, so debugging is no longer possible.

shymega:
> Looking second time into the PR I don't understand the purpose. If we want dmg to be uploaded, we should just do release build and that's it. The build difference between Debug and Release has a significant difference in that Release strips symbols, so debugging is no longer possible.

Yeah, nor do I. I can see the scripts have been stripped of their conditional, as well as the workflow, so that the DMG is built on every commit, just like Windows EXEs, but I'm not too comfortable with the layout so far. And the commits aren't very descriptive, nor was the PR properly created.

I think at this point, seeing as there's been no update or change, and it needs to be rebased against `master`, we should close this PR. I recognise the purpose of having DMGs built on nightlies, but there are better approaches.

shymega:
Closing, we can revisit this if @kukat wants to.

PR #1534, opened 2023-01-06
Use consistent style of license headers in source files

p12tic:
This PR switches license headers to use consistent style. The authorship information has been moved to LICENSE file on the root of the project.

cc @shymega

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,6 @@
-# Barrier -- mouse and keyboard sharing utility
-# Copyright (C) 2018 Debauchee Open Source Group
-# Copyright (C) 2012-2016 Symless Ltd.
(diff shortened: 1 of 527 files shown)


shymega:
@p12tic The only problem I'm seeing in a fair amount of sources in this PR is that authors such as Chris S are being removed and just replaced with 'Input Leap Developers', which I think is a grey area from a legal standpoint. I can't approve this to be merged yet - I'm wondering about the best approach. Also, we need to rebase against master and fix the merge conflicts. Might be best to wait until you're finished with your changes.

p12tic:
I agree that this is a sensitive subject. So I decided it's best to leave it as is.

I opened https://github.com/input-leap/input-leap/pull/1547 to fix a small number of places where adjustments to copyright notice changes were actually needed.

PR #1563, opened 2023-01-10
CI: Remove `brew update` from macOS builds

shymega:
This currently has a breakage, and fails to update, causing the workflow to ultimately fail, and not provide feedback on macOS builds.

This commit removes `brew update`, in an effort to fix CI checks.

## Contributor Checklist:


shymega:
The fix seems to have worked on GH Actions. I expect AP to work fine, as no changes were made there.

p12tic (approved):
Thanks a lot!

PR #1577, opened 2023-01-17
azure-pipelines: Install ninja on macOS

p12tic:
This should speed-up builds significantly


shymega:
This passed tests, I don't think it's worth doing an in-depth review, so I'm gonna go ahead and merge this.

PR #1586, opened 2023-01-19
Use const void* as event target to avoid unnecessary casts

p12tic:


shymega:
@p12tic We should address Ubuntu 18.04 builds before more PRs. I suspect it's the `fail-fast` option causing later builds to fail. Remind me tomorrow, and I'll make a PR.

p12tic:
@shymega The issues are intermittent. The Azure configuration works so we know that the code is correct, thus no need to delay PRs.

PR #1594, opened 2023-01-21
Add libei backend support (and thus ability to run on Wayland)

p12tic:
This is a rebased https://github.com/input-leap/input-leap/pull/1524 which contains support for libei and consequently is able to run on Wayland. Currently this only works on a custom branch of Mutter and thus is not friendly to end users, but at least something works.

libei support is currently disabled by default. It can be enabled by passing `INPUT_LEAP_BUILD_LIBEI=TRUE` to cmake. E.g., if building via `clean-build.sh`: `env B_CMAKE_FLAGS="-DINPUTLEAP_BUILD_LIBEI=TRUE" ./clean_build.sh`.

Testing the functionality requires the following custom packages:

 - mutter https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2628 (tested with 1ffe5c4818)

 - libportal (https://github.com/flatpak/libportal/pull/96 (tested with e1908e7a41)

 - xdg-desktop-portal https://github.com/flatpak/xdg-desktop-portal/pull/714 (tested with c1aa05091e)

 - xdg-desktop-portal-gnome https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/61 (tested with 532889384)
 
 - libei https://gitlab.freedesktop.org/libinput/libei (tested 0.4.1 tag)
 
What currently works: InputLeap client and server, only simple mouse and keyboard actions on Mutter.

What currently does not work:

 - Clipboard sharing.
 
 - Encryption (even though libei backend is not involved with TCP connection between InputLeap server and client)
 
 - The integration with Mutter permission dialogs could be better.
 
 - Task bar icon support.

No release note have been added in this PR. As far as end users are concerned Wayland is still not supported until there is support in upstream mutter.
 
Thanks a lot to @whot and @ofourdan who made this happen!

I want to stress that this particular PR is only a very small part of the puzzle required to get InputLeap to work on Wayland. Going by lines of code this PR is less than 5% of the total amount of changes. @whot and @ofourdan were working in this area for the past several years, without their work this PR couldn't exist.

The original PR description by @whot is below.

----------------

libei is a library for **E**mulated **I**nput and a replacement for what is XTEST under X11. Unlike XTEST it's windowing-system agnostic, it's just a transport layer for sending *logical* input events. See the [libei README](https://gitlab.freedesktop.org/libinput/libei/)  for details on the motivation etc.

libei clients like input-leap can work as receiver context (barriers, real input events are sent to us via libei) or as sender context (barrierc, emulated events are sent by us to the compositor). However, libei is only the transport layer, the logic of establishing a connection is punted to the portals, in particular the `RemoteDesktop` portal for barrierc and the new `InputCapture` portal for barriers.

For the portal communication we're using libportal which requires a GLib mainloop (and thus a separate thread).

This PR adds the required bits to:
- make X11 optional and add libei as a separate backend (`--use-ei` in barrierc/barriers)
- add a GLib mainloop in a separate thread so we can use libportal easily
- add the bits to `barrierc` to talk to [org.freedesktop.portal.RemoteDesktop.ConnectToEIS](https://github.com/flatpak/xdg-desktop-portal/pull/762) to set up libei and then use those devices to send input events
- add the bits to `barriers` to talk to [org.freedesktop.portal.InputCapture](https://github.com/flatpak/xdg-desktop-portal/pull/714) to set up pointer barriers and get input events from the compositor

To actually run this, you will need to replace a bunch of components, see the [tracking issue on libei](https://gitlab.freedesktop.org/libinput/libei/-/issues/1). As of opening the PR, you will need updated mutter, libportal, xdg-desktop-portal, xdg-desktop-portal-gnome and of course libei itself.

The input-leap internal bits are unchanged, so you can use a libei-enabled barriers to talk to input-leap's barrierc on Win/MacOs/X11 and so on.

cc @ofourdan who deserves most of the credit for the hard parts


Lunarequest:
is there any chance this would work on kde plasma? not just gnome?

whot:
Sure, there is nothing binding us to gnome in the patches here and inputleap doesn't even know which compositor is running - the only communication here is with the portal via DBus and then the events via libei. In fact during development I had a python program faking the portal and the eis-demo-server tool from the libei git repo take the place of the compositor.

ghollingworth:
Has anyone tried this with any wlroots based display manager such as Wayfire or Sway?

Do you think there's any point in spending time trying to build for that, or do you think it more likely to be even further broken there?

ofourdan:
> Has anyone tried this with any wlroots based display manager such as Wayfire or Sway?

It needs support in the compositor, wlroots does not have support for libei yet, and also the bits in the [wlroots portal](https://github.com/emersion/xdg-desktop-portal-wlr).

PR #1601, opened 2023-02-22
Fixes X11 clipboard issues, adds size limit for clipboard transfers, added webp/tiff/png/jpg

draekko:
## Contributor Checklist:

* [x] I have created a file in the `doc/newsfragments` directory *IF* it is a
      user-visible change (and make sure to read the `README.md` in that directory) 

The patch was originally made for barrier (a PR was made for it as well) but i only noticed today that it's no longer maintained so adapted the patched code to input-leaf

This fixes X11 to X11 clipboard issues, prevents images from leaking to text targets and other image targets other than the one its intended for. Adds size limits if someone wants to limit clipboard transfers. Adds converters for WEBP, TIFF, PNG, JPG image formats. 

PNG is needed to properly transfer images copy pasted from apps or desktop or from gimp on one machine to the next under X11, added the other formats since it was virtually the same code for them as PNG. The size limit option for the clipboard was mentioned in the barrier wiki but there was no corresponding code, this PR adds the option.

Should not interfere with other platforms but I don't have a Mac or Windows machine for dev support so someone will need to add those there they want it as well. To clarify, only the X11 converters for PNG/WEBP/TIFF/JPG need to be created for the other platforms, the memory limit option for clipboard transfers should work across platforms.

PR #1607, opened 2023-02-25
X11 clipboard image converters

Cimbali:
## Contributor Checklist:

This is a refactor of #1601 splitting commits more semantically.
- 5fc0a9eebbe96047c66decc62afb56e2482c5b2f looks large but it’s mostly new files
- Most `.ui` changes have disappeared as they were just noise (items reordered in the file, but their placement is dictated by the row attribute, not the place in the file). I’ve kept the content changes in 4bf37f3e137baf8377f9df9f4e867e62ddff0df4 -- not sure they are meant to be there or not.
- I’ve also fixed whitespace on the modified lines

### DIFF
--- a/doc/newsfragments/clipboard-sharing-size-limit.feature
+++ b/doc/newsfragments/clipboard-sharing-size-limit.feature
@@ -1,0 +1,1 @@
+Adds a new clipboard sharing size limiter to prevent transferring clipboards past a default value in bytes.
--- a/doc/newsfragments/fix-x11-clipboard.bugfix
+++ b/doc/newsfragments/fix-x11-clipboard.bugfix
@@ -1,0 +1,1 @@
+fixes issue with X11 clipboard sharing images as text targets or in other targetsthan the one it should have.
--- a/doc/newsfragments/x11-new-converters.feature
+++ b/doc/newsfragments/x11-new-converters.feature
@@ -1,0 +1,1 @@
+adds new converters for X11 to support copy pasting png/tiff/jpg/webp formats.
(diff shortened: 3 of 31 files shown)


shymega:
I don't know what's happened to macOS and Ubuntu builds. Before we merge, I want to check with @p12tic that this is normal.

shymega:
Rebased against master, force pull required.

Cimbali:
It seems like a breaking change in the libei API from v0.3 to v0.4, see  [`eis_device_start_emulating()`](https://libinput.pages.freedesktop.org/libei/group__libeis.html#ga7d51d406a1ed01119b1d027f196134f3) spec in the v0.4 docs. This library is only used on ubuntu 22.04 apparently, other builds do not use libei:

https://github.com/input-leap/input-leap/blob/b3ccb7fea1a70b36d91b0196ef836e0b3d2965df/.github/workflows/builds.yml#L108-L120

Cimbali:
Here is [the commit that introduced the extra parameters](https://gitlab.freedesktop.org/libinput/libei/-/commit/ff9830c12228f646174711a5d4b12b6a51b52b16), it’s 2 weeks old. The problem only happens because the CI is building from git.

shymega:
Fixed in #1611. I'll rebase this PR against mainline, and we can just check the CI definitely all passes. You may need to force pull.

Cimbali:
My bad actually it was a post-v0.4.1 change, so you could pin to 0.4.1. The version displayed on their docs site confused me, they display the current master docs with the last-release version, even though these are not in sync in that case.

shymega:
> My bad actually it was a post-v0.4.1 change, so you could pin to 0.4.1. The version displayed on their docs site confused me, they display the current master docs with the last-release version, even though these are not in sync in that case.

Oh, right, I did originally have a branch for 0.4.1, but I can push a bump from 0.3 to 0.4.1. Didn't realise. Anyway, your PR passed checks, so I'll merge now.

Cheers!

PR #1610, opened 2023-02-28
CI (libei): Lock clone of libei to 0.4.1

shymega:
This should fix builds where cloning from Git mainline, causes a breakage in CI.

## Contributor Checklist:


shymega:
Closing, targeted the wrong tag - we should target 0.3 for now, as that works, and 0.4 introduced the breaking change AFAICT.

PR #1617, opened 2023-03-03
Build fixes

Cimbali:
## Contributor Checklist:

----

These are code improvements only, mostly upgrading from deprecated code, so no user-visible change.

1. 4628a7c8b98064f58a5f4dc64cc5680adb12e68a fixes building against the last libportal release (currently CI builds against master, which has caused issues when done with libei)
2. 3557fb55297b011c51fc42a9174a3f880d5d6523 fixes building against the latest libei (i.e. master), which will inevitably be needed at some point
3. 0497e17ba9b18e0cf941733ae56281bdc7fd15a3 is a bug fix, a return statement was missing
4. 6edd9363ac4444b2305585af568e30333b8a46bf a5ff6ced6b797340287cf39aedbd0e10d22696be b4cb8339d4712627f92b53b663b5aad35cb584f4 are 3 different Qt deprecations. Useful to start fixing code before upgrading to Qt 6, this will allow to minimize changes.
7. 9e72ce97414e824946e733f2172ba3eb5f74618b fixes the recommended way of creating an RSA key, the current used way is deprecated since 0.9.8 (latest stable is 3.0 and introduces this new shorthand function, it’s also LTS until Sept. 2026).
8. c88ec801181c4106535bc27db585f275a92d4819 removes 2 `std::move()` calls that are redundant or prevent compiler optimisations.

All deprecation fixes are done in a way that does not break current builds or change library requirements.

The remaining fixes were in the libei-only part of the code:

9. 8e4cf4957d0ea706c57ad328c4850793c9654441 fix initialisation order. Slightly nitpicky but since I was fixing warnings anyway, I might as well.
10. 86fd01ae913ed1b64814d3f53faccc45308d7441 a `break` was missing between device paused and device resumed events, which would cause both events to be logged on a pause event.
11. 07be35f1a62967c909149301cfb0e01ade4e742e Initialise a struct correctly. If using reclaimed memory (which typically happens on the stack) the pointers in the incorrectly initialised struct will contain garbage and risk causing segfaults.

If you compile with warning flags after this PR, the only remaining warnings are unused parameters, i.e. `-Wall -Wextra -Wno-unused-parameter` will output no warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,11 +156,16 @@ if (UNIX)
             # no libportal release has xdp_session_connect_to_eis, so let's check for that
             # and enable the hackaround in the code
             include(CMakePushCheckState)
+            include(CheckCXXSourceCompiles)
             cmake_push_check_state(RESET)
             set(CMAKE_REQUIRED_INCLUDES "${CMAKE_REQUIRED_INCLUDES};${LIBPORTAL_INCLUDE_DIRS};${GLIB2_INCLUDE_DIRS}")
             set(CMAKE_REQUIRED_LIBRARIES "${CMAKE_REQUIRED_LIBRARIES};${LIBPORTAL_LINK_LIBRARIES};${GLIB2_LINK_LIBRARIES}")
             check_symbol_exists(xdp_session_connect_to_eis "libportal/portal.h" HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)
             check_symbol_exists(xdp_input_capture_session_connect_to_eis "libportal/inputcapture.h" HAVE_LIBPORTAL_INPUTCAPTURE)
+            # check_symbol_exists can’t check for enum values
+            check_cxx_source_compiles("#include <libportal/portal.h>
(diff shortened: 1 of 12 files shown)


p12tic:
Thanks a lot for the PR.

Everything looks good, except for the `endl` changes, can we instead replace endl with `"\n"`? endl is "\n" followed by flush and in these cases we don't need to flush the file.

Cimbali:
Rebased on master + added requested changes (using `"\n"` instead of `Qt::endl`).

shymega:
There's no newsfragment in this PR, so do we need the checkbox ticked?

p12tic (approved):

Cimbali:
> There's no newsfragment in this PR, so do we need the checkbox ticked?

I understood the prompt as “I checked whether a newsfragment was needed, and added one in the case it was” -- I guess it’s semantics at this point.

shymega:
It's more, you tick the checkbox if you *have* created a newsfragment. I've noticed a fair few people ticking the box even if there's no newsfragment. Perhaps we should rephrase the question.

PR #1621, opened 2023-03-07
AppData validation fix

aekoroglu:
According to our packaging policy (https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/) AppData files must correctly validate using appstream-util validate-relax.


shymega:
This is great, merging, thank you!

PR #1632, opened 2023-04-03
Fix further OpenSSL deprecation warnings

Cimbali:
- On OpenSSL ≥ 3.0 we use the up-to-date code. In versions ≥ 0.9.8 but < 3.0 we get a warning. Unfortunaltely some (CI) platforms still do not have access to OpenSSL ≥ 3.0
- `EVP_PKEY_assign_RSA()` was also deprecated since ≥ 3.0 since it is not needed anymore: EVP_RSA_gen() returns a EVP_PKEY* directly. Remove the EVP_PKEY_assign_RSA() call and use explicit pointers instead of auto to clarify what is going on.

I initially added deprecation ignores for OpenSSL ≥ 0.9.8 < 3.0 but adding the proper code is not that much harder if we need to specifically cater to that series of OpenSSL versions anyway (with warning ignores or other).

## Contributor Checklist:

### DIFF
--- a/src/lib/net/SecureUtils.cpp
+++ b/src/lib/net/SecureUtils.cpp
@@ -161,22 +161,33 @@ FingerprintData get_pem_file_cert_fingerprint(const std::string& path, Fingerpri
 void generate_pem_self_signed_cert(const std::string& path)
 {
     auto expiration_days = 365;
+    constexpr unsigned key_bits = 2048;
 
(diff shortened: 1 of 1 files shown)


p12tic:
Looks good, I only had a couple of nitpicks.

shymega:
Can we merge this now @p12tic?

PR #1635, opened 2023-04-07
macOS: add support for right side modifier keys

Koston-0xDEADBEEF:
Notes:
 * Tested with FreeBSD server & macOS client
 * Function `handleModifierKeys()` is untested; I was unable to trigger a breakpoint in it. Maybe it's only hit when running a server?
 
 There's still something a bit off in the logic when handling multiple simultaneous modifier key presses: Pressing Alt_R-Shift_L works, but pressing Shift_L-Alt_R causes latter to get interpreted with incorrect keycode. I'm not sure if I have time/energy left to address it, but I'll try.
Edit: This seems to be an unrelated XWindow server side issue.
Edit2: It's not a bug, just what happens when shift modifier is applied to alt_r with my current configuration

### DIFF
--- a/src/lib/platform/OSXKeyState.cpp
+++ b/src/lib/platform/OSXKeyState.cpp
@@ -28,17 +28,6 @@
 
 namespace inputleap {
 
-// Note that some virtual keys codes appear more than once.  The
-// first instance of a virtual key code maps to the KeyID that we
-// want to generate for that code.  The others are for mapping
-// different KeyIDs to a single key code.
-static const std::uint32_t s_shiftVK = kVK_Shift;
-static const std::uint32_t s_controlVK = kVK_Control;
-static const std::uint32_t s_altVK = kVK_Option;
-static const std::uint32_t s_superVK = kVK_Command;
-static const std::uint32_t s_capsLockVK = kVK_CapsLock;
-static const std::uint32_t s_numLockVK = kVK_ANSI_KeypadClear; // 71
-
 static const std::uint32_t s_brightnessUp = 144;
 static const std::uint32_t s_brightnessDown = 145;
 static const std::uint32_t s_missionControlVK = 160;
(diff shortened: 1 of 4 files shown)


Koston-0xDEADBEEF:
Code is now tested both as client and server and added support for distinguishing right side modifiers also as server.

src/lib/platform/OSXKeyState.cpp
```diff
@@ -108,22 +97,21 @@ static const KeyEntry    s_controlKeys[] = {
...
-    { kKeyControl_L,    s_controlVK },
-    { kKeyControl_R,    s_controlVK }, // 62
-    { kKeyAlt_L,        s_altVK },
-    { kKeyAlt_R,        s_altVK },
-    { kKeySuper_L,        s_superVK },
-    { kKeySuper_R,        s_superVK }, // 61
-    { kKeyMeta_L,        s_superVK },
-    { kKeyMeta_R,        s_superVK }, // 61
+    // modifier keys. Support for right side keys is a later addition in macOS.
+    { kKeyShift_L,          kVK_Shift },        // 56
```
p12tic, line 101: Let's align with the comma (`{ kKeyShift_L, kVK_Shift },`), because otherwise all lines will need to be modified whenever a very long first work appears in the list.

p12tic:
@Koston-0xDEADBEEF Thanks a lot for the PR. It looks good, I only had a single nitpick regarding alignment.

src/lib/platform/OSXKeyState.cpp
Koston-0xDEADBEEF: fixed

Koston-0xDEADBEEF:
> @Koston-0xDEADBEEF Thanks a lot for the PR. It looks good, I only had a single nitpick regarding alignment.

Cheers - I fixed the alignment as requested.

shymega:
@Koston-0xDEADBEEF All looks good to me. One more thing, could you rebase your PR to squash the two commits into one? That way we don't have a dangling 'fix alignment' commit. Once that's done, either @p12tic or I can merge. Cheers.

Koston-0xDEADBEEF:
> @Koston-0xDEADBEEF All looks good to me. One more thing, could you rebase your PR to squash the two commits into one? That way we don't have a dangling 'fix alignment' commit. Once that's done, either @p12tic or I can merge. Cheers.

I've squashed the commits and rebased against upstream.

shymega:
OK, I'll merge it now.

shymega:
Thank you for your work on this, @Koston-0xDEADBEEF, very grateful for your effort. :)

Koston-0xDEADBEEF:
> Thank you for your work on this, @Koston-0xDEADBEEF, very grateful for your effort. :)

Happy to help. I will see if I can also fix some memory corruption bugs while at it.

PR #1636, opened 2023-04-10
[fix] fix server hangs on mac suspend

gdh1995:
This PR fix a dead lock that `EventQueue::deleteTimer` may lock its `mutex_` twice.

On my macOS 13, `input-leaps` always hangs if I close the lid when a ubuntu client is connected, and after some tries I found this bug. The 2nd `lock_guard` constructed during `delete timer` will enter a dead loop, making the whole process hang until a `kill -9` command.

This PR ensures the `mutex_` has been released before `delete timer`, so the bug disappears. Maybe a `recursive_mutex` will work better, but this PR is also enough.

### DIFF
--- a/doc/newsfragments/1636_fix_hang_on_macos_suspend.bugfix
+++ b/doc/newsfragments/1636_fix_hang_on_macos_suspend.bugfix
@@ -1,0 +1,1 @@
+Fix a dead lock entered when a server screen is suspended, so that screen-resume message will work as expected.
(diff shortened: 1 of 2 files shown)


p12tic (approved):
Good catch, thanks.

p12tic:
@gdh1995 Could you add a release note in doc/newsfragment directory? doc/newsfragment/README.md explains how to do so and there are a lot of examples there already.

gdh1995:
Hello, I've added 1636_fix_hang_on_macos_suspend.bugfix

gdh1995:
Oh now rebased

shymega:
Merged. Thank you!

PR #1624, opened 2023-03-21
clipboard: block updates when sender is not the owner

vancluever:
This blocks clipboard updates when the sender is not the clipboard owner.

This is actually being done in an assertion in onClipboardChanged, but obviously these are removed from production builds. As such, this is effectively allowing clipboard updates from any screen, no matter if it's the current screen or not, which breaks the clipboard state.  This can cause crashes of the server if the screen is moved at certain times during this process.

Note that this has a side-effect of possibly blocking clipboard updates if the screen is switched before a clipboard update is acknowledged. This delay is obviously quite short, so observing it is hopefully rare.

Fixes #1593.

## Contributor Checklist:

* [x] I have created a file in the `doc/newsfragments` directory *IF* it is a
      user-visible change (and make sure to read the `README.md` in that directory)

Issue #1433, opened 2022-01-15
Clipboard being updated by non-active screen


soraxas:
### What happened?

Would it be possible to add an option to disable clipboard sharing across devices?

The reason is that I think it currently conflicts with other clipboard sharing applications. I have been using [Join](https://joaoapps.com/join/) which include sharing clipboard across clients (across devices such as PC/laptops/android phone/etc.).

Currently while using barrier (or the to-be input-leap), using PC1 as host, copying and pasting on client (PC2) will somehow erase the content. The following are logs from Host (while operating on Client):
```
[2022-01-15T14:42:47] INFO: screen "Client" grabbed clipboard 0 from "Host"
[2022-01-15T14:42:47] INFO: screen "Client" grabbed clipboard 1 from "Host"
# I think at this point *Join* had sent a network request to update clipboard on PC1 (Host). I haven't done any other operations at this point
[2022-01-15T14:42:50] INFO: screen "Host" grabbed clipboard 0 from "Client"
[2022-01-15T14:42:50] INFO: screen "Host" grabbed clipboard 1 from "Client"
```
where the lines
```
[2022-01-15T14:42:50] INFO: screen "Host" grabbed clipboard 0 from "Client"
[2022-01-15T14:42:50] INFO: screen "Host" grabbed clipboard 1 from "Client"
```
ended up resetting the clipboard.

### Version

v2.4.0

### If applicable, where did you install Barrier from?

Host: Linux by compiling
Client: Windows from release

### What OSes are you seeing the problem on? (Check all that apply)

Linux, Windows

### What OS versions are you using?

Host: Arch Linux with latest Kernel
Client: Windows 10

albertony:
There is an option in GUI: Server Configuration > Advanced server settings > Enable clipboard sharing.
Is this not covering what you want? I have not tested it myself, so its an open question..

soraxas:
Oh! I haven't even realised there are options there (I've only been looking at `Barrier > Settings`). Thanks for pointing that out. It currently works as what I had in my mind. Btw I recalled seeing some post about making Barrier server-client agnostic (keyboard/mouse works both way). When that happens, perhaps these settings should be moved to the general location?

However, that sparks my other question of shouldn't the clipboard grabbing only happens on the current active screen? Say, current barrier currently keep track of the **active** screen:
```
[2022-01-16T10:04:09] INFO: switch from "Client" to "Host" at 1887,659
[2022-01-16T10:04:09] INFO: entering screen
[...]

PR #1640, opened 2023-04-29
Revert #1433 (blocking off-screen clipboard updates)

vancluever:
This reverts #1433.

Based on discussion in #1624 it seems like this would be better denoted as intended behavior and if a truly shared clipboard was not desired, that clipboard sharing should be disabled.

Blocking off-screen updates truly is a complex process and there are other parts of the stack that need to be fixed as well (see aforementioned issue). If this is a feature that is desired, a larger update will be needed, and likely should also be introduced as an optional feature.

## Contributor Checklist:


shymega:
Can you squash all three commits into one? I can't review this myself, have assigned Povilas. But for now, squashing would be helpful.

vancluever:
@shymega done!

shymega:
Thanks! :+1:

shymega:
@p12tic Any updates on review of this PR?

p12tic (approved):

PR #1646, opened 2023-05-14
fix: wake up on mac when screen locked

shymega:
@shymega's note: from debauchee/barrier#1927, thank you @junbujianwpl.

I'm creating PRs from Barrier's unmerged PRs for Input Leap. This is so we can a) let the community know that there is an active fork and b) take advantage of the code contributions that Barrier simply isn't accepting. This gives us the benefit of improved codebase and bug fixes, such as this one, which happens on Mac machines.

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/OSXScreen.mm
+++ b/src/lib/platform/OSXScreen.mm
@@ -811,6 +811,8 @@ OSXScreen::enter()
 			IORegistryEntrySetCFProperty(entry, CFSTR("IORequestIdle"), kCFBooleanFalse);
 			IOObjectRelease(entry);
 		}
+		IOPMAssertionID assertionID;
+		IOPMAssertionDeclareUserActivity(CFSTR("Input Leap Wakeup Assertion"), kIOPMUserActiveLocal, &assertionID);
 
 		avoidSupression();
 	}


src/lib/platform/OSXScreen.mm
```diff
@@ -811,6 +811,8 @@
 			IORegistryEntrySetCFProperty(entry, CFSTR("IORequestIdle"), kCFBooleanFalse);
 			IOObjectRelease(entry);
 		}
+		IOPMAssertionID assertionID; 
+		IOPMAssertionDeclareUserActivity(CFSTR(""), kIOPMUserActiveLocal, &assertionID);
```
p12tic, line 815: According to the documentation we should fill in the AssertionName field (first parameter).

src/lib/platform/OSXScreen.mm
```diff
@@ -811,6 +811,8 @@
 			IORegistryEntrySetCFProperty(entry, CFSTR("IORequestIdle"), kCFBooleanFalse);
 			IOObjectRelease(entry);
 		}
+		IOPMAssertionID assertionID;
```
Koston-0xDEADBEEF, line 814: nitpick, extra whitespace at end of line

Koston-0xDEADBEEF:
Tested, works. macOS client now wakes up faster and by just moving mouse over, without patch keyboard input is required.

src/lib/platform/OSXScreen.mm
shymega: Resolved in latest force-push and rebase.

src/lib/platform/OSXScreen.mm
shymega: What name would you suggest? Or should we just put something like 'INPUT-LEAP_WAKE'?

p12tic (approved):

p12tic:
LGTM.

PR #1648, opened 2023-05-15
Fix scrolling speed on macOS client

Koston-0xDEADBEEF:
There was an ad-hoc fix in place for matching scrolling speed with macOS, which is no longer needed.


p12tic:
I agree with @Koston-0xDEADBEEF assessment, we should have a GUI option for this so that the users can decide for themselves which way they prefer better.

shymega:
Concur.

shymega:
I'll address this later on tomorrow, but I think I would have rathered for a rebase, than a merge. If we merge this commit, we would have a merge commit in the PR, when it would've been a bit cleaner for a rebase.

What I'll do, is I'll revert the merge, rebase, and then force push. The original commit will remain, but it's cleaner this way.

Koston-0xDEADBEEF:
> I'll address this later on tomorrow, but I think I would have rathered for a rebase, than a merge. If we merge this commit, we would have a merge commit in the PR, when it would've been a bit cleaner for a rebase.
> 
> What I'll do, is I'll revert the merge, rebase, and then force push. The original commit will remain, but it's cleaner this way.

I agree with rebasing, but rather than merging this change at all, it should be extended to make the scrolling speed at minimum a compile-time option, preferably a runtime configuration option. This is on my TODO-list; apologies for not being clear about it.

shymega:
> > I'll address this later on tomorrow, but I think I would have rathered for a rebase, than a merge. If we merge this commit, we would have a merge commit in the PR, when it would've been a bit cleaner for a rebase.
> > What I'll do, is I'll revert the merge, rebase, and then force push. The original commit will remain, but it's cleaner this way.
> 
> I agree with rebasing, but rather than merging this change at all, it should be extended to make the scrolling speed at minimum a compile-time option, preferably a runtime configuration option. This is on my TODO-list; apologies for not being clear about it.

Yeah, I haven't been able to revert the merge commit yet.

Would you like to approach this problem, by closing this PR, and opening a new PR with a configuration option? I think runtime would be best.

Cheers.

Koston-0xDEADBEEF:
Sure thing. Sorry for the delay, I'll do a new pull request with runtime conf option when I have some spare time.

shymega:
No worries. Thanks for closing the PR for me.

PR #1649, opened 2023-05-17
ei: adjust to API changes in libei 1.0.0rc1

whot:
libei has a libei-1.0 pkconfig file now, so let's update to that. The rest of the APIs are fairly straightforward renames with the exception of the new BUTTON and SCROLL capabilities (previously part of POINTER/POINTER_ABSOLUTE).

Technically it's possible for an EIS implementation to create a pointer device without buttons or scroll but those wouldn't work well with our implementation so let's just ignore those and require button + scroll.

## Contributor Checklist:


whot:
Side-note; the libei protocol/API are now stable

p12tic (approved):

PR #1654, opened 2023-05-24
base: only remove handlers if we actually have a client

whot:
This caused a null-pointer dereference in EventQueue::remove_handler since that was unconditionally called even when the client was NULL.

## Contributor Checklist:


shymega:
Good catch @whot, thanks - CI passed well, merging.

PR #1658, opened 2023-06-08
CI: bump to build against libei 1.0.0

whot:
Also move this into an env, this way it sticks out a bit more

There was a last-minute minor libeis ABI and protocol change though neither of those would have any effect on the bits input-leap uses. Still, better to build against the latest released version.

## Contributor Checklist:

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -81,17 +81,19 @@ jobs:
         run: git config --global --add safe.directory $GITHUB_WORKSPACE
 
       - if: matrix.os == 'ubuntu:22.04'
-        name: build libei from git tag (0.99.1)
+        name: build libei from git tag (1.0.0)
         run: |
             apt-get install -y python3-pip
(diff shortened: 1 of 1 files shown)


shymega:
I think we can squash the contents of this PR into the new 'Update CI' PR upcoming. Waiting for @p12tic.

whot:
> I think we can squash the contents of this PR

wfm and FTR if you ever need to squash one of my commits elsewhere for easier maintenance, feel free to do so, no need to ask.

shymega:
Thanks @whot. I don't know if you can push to that PR, but I can cherry-pick this PR's commit into the 'Update CI' PR, and we can close this PR. Does that work for you?

shymega:
OK, so cherry-picking doesn't apply cleanly. I'll have to do this manually.

shymega:
I've pushed to that PR manually. I will now close this PR - thank you for the deps bump.

Issue #1108, opened 2021-11-04
Edge detection fails with DPI other than 100%


ghost:
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/1108
> 
> Issue created on: 2021-03-25 by @oilcan-productions
> Issue last updated on: 2021-05-26

My Server is Windows 10 20H2 (does not really matter was the same issue before)
My Server has 2 monitors
M1 = Laptop Display (Main)
M2 = Display Port Monitor

I set my DPI on both to 150% (Recommended) 

My clients are Windows 10 20H2, Raspberry Pi, Mac OS Big Sur (latest)

the config is 
``` 
            RaspberryPi
                 |
MacOS  -- Server M1 -- Server M2 -- Win10
```

With DPI set to 150% on the Server I can move from M1 to MacOS or RaspberryPi
From M2 I cannot move to Win10
When I set the DPI to 100% on both displays of the Server I can move between all displays successfully.
When I move Win10 Below M1 I can move to it regardless of the DPI settings
From the log it seems that with DPI != 100% the edge of M2 is not detected at all as I don't see the event in the logs.

---

> Commented on: 2021-03-30 by @verglor

I have similar problem with 3 monitors in row on my Win10 server and one macOS Catalina client below them.
If all my server monitors have the same DPI everything works fine. Once they have different DPI from each other, I cannot move to macOS client via edge however I can move from client to server.

---

> Commented on: 2021-05-26 by @neakmenter

I have the same issue here. main 4k screen is 250% scaling, 1080 screen 'above' and 1080 screen on the left both at 100%. want to "barrier" over to surface pro 3 screen on the right. It works if I place both my 1080 screens to the left of the main screen, or if I disconnect one of them. But will not work with my usual setup unless I reduce scaling to less than 150% on the main screen. I've been using a keyboard shortcut to get to the surface screen, but it's far less intuitive. (the cursor will 'return' to the server just fine, just can break through to the client!)
Just tried something - it works if I arrange my monitors like this screen...
![image](https://user-images.githubusercontent.com/69205639/119714955-5ad2d500-be5b-11eb-81b7-eb6a82549441.png)

futurizer:
same issue here as well. main 4k screen 30inch screen with 150% DPI setting on Windows ( as server , left ) and one another intel macOS Monterey ( as client , right )

version on windows
[...]

PR #1669, opened 2023-06-30
Re-add manifests for Windows with DPI-aware flag

lreeves:
Seems like a long time ago when this was named Barrier these files got removed here:

https://github.com/input-leap/input-leap/commit/aea488167af19845369e062bf9af4edfadc4fa71

Adding them back into the Windows build seems to correct the issues people have been having moving the mouse away from Windows servers that have their DPI scaling set to anything other than 100%. Should fix these Github issues:

- https://github.com/debauchee/barrier/issues/1713
- https://github.com/input-leap/input-leap/issues/1108

### DIFF
--- a/src/client/CMakeLists.txt
+++ b/src/client/CMakeLists.txt
@@ -21,7 +21,10 @@ set(sources
 if (BUILD_MSWINDOWS)
     file(GLOB mswin_headers "MSWindows*.h")
     file(GLOB mswin_sources "MSWindows*.cpp")
-    list(APPEND sources input-leapc.rc)
+    list(APPEND sources
+        input-leapc.exe.manifest
+        input-leapc.rc
+    )
     set_property(SOURCE input-leapc.rc PROPERTY COMPILE_FLAGS /nologo)
 endif()
 if (BUILD_CARBON)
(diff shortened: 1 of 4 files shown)


xoascf:
Since a manifest is being added, it would be a good idea to [enable visual styles](https://learn.microsoft.com/en-us/windows/win32/controls/cookbook-overview) as well.

shymega:
Good catch @lreeves.

@xoascf Can you create a separate issue for this PR - I would rather keep PRs to one feature/enhancement/bugfix. This ensures we can revert if we need to.

lreeves:
@shymega anything else I should do to get this merged?

shymega:
> @shymega anything else I should do to get this merged?

No, but I want it reviewed before I can merge.

ViNull008:
works for me! now I could use my windows laptop as server and contorl my Ubuntu PC.
I'm goint to give up Barrier/input-leap until I got this building.
Thanks to @lreeves and @shymega

shymega:
Good to hear.

shymega:
Merging... 

Checks passed, and I see no reason to keep this pending any longer.

PR #1687, opened 2023-09-13
RFC: When portal code is builtin silence Wayland support warning

prahal:
This is a user visible in that is disable a warning popup if running under wayland.
I did not add a doc/newfragments as there were none when introducing this popup.

This is an RFC as we might want to keep the popup to tell the user that wayland is still officially not supported.
Still, I find that my code only removes the popup if the EI support is building which is disabled by default. So it is redundant to me.
The popup is kept for if the libei/inputcapture support is not builtin and running under a wayland session.

I wonder if config.h should be imported in a header instead of the implementation as I did.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,12 +173,6 @@ if (UNIX)
             install(FILES "dist/flatpak/input-leap-flatpak"
                     DESTINATION bin
                     PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
-
-	    if (INPUTLEAP_WARN_ON_WAYLAND)
-                add_definitions(-DINPUTLEAP_WARN_ON_WAYLAND=1)
-                set(INPUTLEAP_WARN_ON_WAYLAND=1)
-            endif()
-
         endif()
 
         check_include_files ("dns_sd.h" HAVE_DNSSD)
@@ -207,6 +201,10 @@ if (UNIX)
         if(INPUTLEAP_BUILD_LIBEI)
             add_definitions(-DWINAPI_LIBEI=1)
             set(BUILD_LIBEI 1)
+            if (INPUTLEAP_WARN_ON_WAYLAND)
+                add_definitions(-DINPUTLEAP_WARN_ON_WAYLAND=1)
+                set(INPUTLEAP_WARN_ON_WAYLAND 1)
+            endif()
         endif()
         if(NOT INPUTLEAP_BUILD_X11 AND NOT INPUTLEAP_BUILD_LIBEI)
             message(FATAL_ERROR "One of X11 or libei is required")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(INPUTLEAP_BUILD_TESTS "Build the tests" ON)
 option(INPUTLEAP_USE_EXTERNAL_GTEST "Use external installation of Google Test framework" OFF)
 option(INPUTLEAP_BUILD_X11 "Build with XWindows support" ON)
 option(INPUTLEAP_BUILD_LIBEI "Build with libei support" OFF)
+option(INPUTLEAP_WARN_ON_WAYLAND "Warn the user about Wayland support when run on Wayland" ON)
 
 set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set (CMAKE_CXX_STANDARD 14)
@@ -172,6 +173,12 @@ if (UNIX)
             install(FILES "dist/flatpak/input-leap-flatpak"
                     DESTINATION bin
                     PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
+	    if (INPUTLEAP_WARN_ON_WAYLAND)
+                add_definitions(-DINPUTLEAP_WARN_ON_WAYLAND=1)
+                set(INPUTLEAP_WARN_ON_WAYLAND=1)
+            endif()
+
         endif()
 
         check_include_files ("dns_sd.h" HAVE_DNSSD)
--- a/src/client/input-leapc.cpp
+++ b/src/client/input-leapc.cpp
@@ -20,6 +20,7 @@
 #include "arch/Arch.h"
 #include "base/Log.h"
 #include "base/EventQueue.h"
+#include "config.h"
 
 #if WINAPI_MSWINDOWS
 #include "MSWindowsClientTaskBarReceiver.h"
@@ -31,8 +32,7 @@ namespace inputleap {
 CreateTaskBarReceiverFunc createTaskBarReceiver = nullptr;
 #endif
 
-int client_main(int argc, char** argv)
-{
+int client_main(int argc, char** argv) {
 #if SYSAPI_WIN32
     // record window instance for tray icon, etc
     ArchMiscWindows::setInstanceWin32(GetModuleHandle(nullptr));
@@ -43,6 +43,22 @@ int client_main(int argc, char** argv)
 
     Log log;
     EventQueue events;
+
+    // TODO: Remove once Wayland support is stabilised.
+
+    // This block only warns when `libportal` and `libeis` aren't available - as well as if the top-level CMake option is enabled - by default, it is.
+    // It serves as a way to warn users that using Wayland on platforms matching this boolean expression, that their platform might not be fully supported.
+    // It does *not* run on X11 platforms, Win32, or macOS.
+#if (defined(WINAPI_LIBEI) && defined(INPUTLEAP_WARN_ON_WAYLAND)) && (!defined(HAVE_LIBPORTAL_INPUTCAPTURE) || !defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS))
+    const char *val = std::getenv("WAYLAND_DISPLAY");
+    if (val == nullptr) {
+        // Return, we're not running on Wayland. Possibly. Could be enhanced.
+        return
+    } else {
+        // We are running on Wayland.
+        // TODO: Add log message.
+    };
+#endif
 
     ClientApp app(&events, createTaskBarReceiver);
     int result = app.run(argc, argv);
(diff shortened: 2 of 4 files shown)


shymega:
I mean, I'm not against this PR. I'm just unsure about including `config.h`, instead of just using `#if defined()`. I don't want more includes than necessary.

It would also be good to have the check in the CLI interfaces too. Not everyone uses the GUI.

prahal:
> I'm just unsure about including `config.h`, instead of just using `#if defined()`. I don't want more includes than necessary.

I do not understand how to do it other than including config.h. My doubts were about where to include the config.h.

`#if defined()` cannot test a constant is defined if this constant is defined in the config.h and the config.h is not included, or do I miss something? (mind the constants I test are only defined in config.h).

Agreed that it would be good to expand the warning to the log and stderr. However, this is another issue (I mean it was already one before this PR).

shymega:
On Wed, Sep 13, 2023 at 09:41:34AM -0700, Alban Browaeys wrote:
> > I'm just unsure about including `config.h`, instead of just using `#if defined()`. I don't want more includes than necessary.
> 
> I do not understand how to do it other than including config.h. My doubts were about where to include the config.h.
> 
> `#if defined()` cannot test a constant is defined if this constant is defined in the config.h and the config.h is not included, or do I miss something? (mind the constants I test are only defined in config.h).
> 
> Agreed that it would be good to expand the warning to the log and stderr. However, this is another issue (I mean it was already one before this PR).

I've made a commit to fix the issues I found. I've also added a CMake option for
controlling this functionality.

The `config.h` is included elsewhere in Input Leap, which the `main.cpp`
inherits.

The `#if defined()` change I made works for me in my tests, so that's all sorted.

I've attached it to this email - if you want to apply the patch and push to your
PR, please go ahead. Or I can do it :)

Best wishes,
--
Dom Rodriguez
GPG Fingerprint: EB0D 45E6 D0DC 1BA1 A2B5  FC24 72DC F123 1E54 BD43

From b084d18fc8d7161fdc2bdf719c229a3afe7e2c6b Mon Sep 17 00:00:00 2001
From: Dom Rodriguez ***@***.***>
Date: Wed, 13 Sep 2023 18:55:33 +0100
Subject: [PATCH] Allow Wayland warning to be disabled/enabled at compile-time

This commit creates a CMake 'option' for disabling/enabling the Wayland
warning. It is enabled only under Unix, non-Apple targets, and by
default, the warning is enabled.

The warning is *also* controlled by the 'libei' definitions created
during compile-time.

The reason for this commit was to allow users to override the otherwise
fine, automatic conditionals detecting Wayland support.

Signed-off-by: Dom Rodriguez ***@***.***>
---
 CMakeLists.txt       | 4 ++++
 src/gui/src/main.cpp | 4 +---
 2 files changed, 5 insertions(+), 3 deletions(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index a6ca11cd..8a34809d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(INPUTLEAP_BUILD_TESTS "Build the tests" ON)
 option(INPUTLEAP_USE_EXTERNAL_GTEST "Use external installation of Google Test framework" OFF)
 option(INPUTLEAP_BUILD_X11 "Build with XWindows support" ON)
 option(INPUTLEAP_BUILD_LIBEI "Build with libei support" OFF)
+option(INPUTLEAP_WARN_ON_WAYLAND "Warn the user about Wayland support when run on Wayland" ON)
 
 set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set (CMAKE_CXX_STANDARD 14)
@@ -204,6 +205,9 @@ if (UNIX)
         if(NOT INPUTLEAP_BUILD_X11 AND NOT INPUTLEAP_BUILD_LIBEI)
             message(FATAL_ERROR "One of X11 or libei is required")
         endif()
+        if (INPUTLEAP_WARN_ON_WAYLAND)
+            add_definitions(-DINPUTLEAP_WARN_ON_WAYLAND)
+        endif()
     endif()
 
 elseif (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
diff --git a/src/gui/src/main.cpp b/src/gui/src/main.cpp
index edd49fb8..43bad308 100644
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -38,8 +38,6 @@
 #include <cstdlib>
 #endif
 
-#include "config.h"
-
 class QThreadImpl : public QThread
 {
 public:
@@ -108,7 +106,7 @@ int main(int argc, char* argv[])
 	QApplication::setQuitOnLastWindowClosed(false);
 
     // TODO: Remove once Wayland support is stabilised.
-#if !HAVE_LIBPORTAL_INPUTCAPTURE or !HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS
+#if (!defined(HAVE_LIBPORTAL_INPUTCAPTURE) || !defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)) && defined(INPUTLEAP_WARN_ON_WAYLAND)
     if (QGuiApplication::platformName() == "wayland") {
         QMessageBox::warning(
         nullptr, "InputLeap",
-- 
2.41.0

shymega:
OK, so I can see that attachment didn't work. Darn!

Would you like me to apply my patch directly to your PR? I've marked it as 'draft' for now.

prahal:
@shymega you could apply a squashed version of the set.

But it seems it does not work for me:
```
grep -E "(WARN_|HAVE_LIBPORT)" CMakeCache.txt 
INPUTLEAP_WARN_ON_WAYLAND:BOOL=ON
HAVE_LIBPORTAL_INPUTCAPTURE:INTERNAL=1
//Test HAVE_LIBPORTAL_OUTPUT_NONE
HAVE_LIBPORTAL_OUTPUT_NONE:INTERNAL=1
HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS:INTERNAL=1

grep -E "(WARN_|HAVE_LIBPORT)" src/lib/config.h 
#define HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS 1
#define HAVE_LIBPORTAL_INPUTCAPTURE 1
```
and I get the warning popup.

I create a build directory, change to it, then:
```
cmake ../ -DINPUTLEAP_BUILD_LIBEI=1
make -j4
./bin/input-leap
```

to verify I deleted my previous build directory and started afresh.

prahal:
> It would also be good to have the check in the CLI interfaces too. Not everyone uses the GUI.

Did I understand correctly. You mean adding a warning about "wayland support" to stderr? (and input-leap log?)

shymega:
> @shymega you could apply a squashed version of the set.
> 
> But it seems it does not work for me:
> 
> ```
> grep -E "(WARN_|HAVE_LIBPORT)" CMakeCache.txt 
> INPUTLEAP_WARN_ON_WAYLAND:BOOL=ON
> HAVE_LIBPORTAL_INPUTCAPTURE:INTERNAL=1
> //Test HAVE_LIBPORTAL_OUTPUT_NONE
> HAVE_LIBPORTAL_OUTPUT_NONE:INTERNAL=1
> HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS:INTERNAL=1
> 
> grep -E "(WARN_|HAVE_LIBPORT)" src/lib/config.h 
> #define HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS 1
> #define HAVE_LIBPORTAL_INPUTCAPTURE 1
> ```
> 
> and I get the warning popup.
> 
> I create a build directory, change to it, then:
> 
> ```
> cmake ../ -DINPUTLEAP_BUILD_LIBEI=1
> make -j4
> ./bin/input-leap
> ```
> 
> to verify I deleted my previous build directory and started afresh.
> 
> If I add: `#define HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS 3↲` in `src/gui/src/main.cpp`, I get no duplicate definition error when running `make`.

I've just done that. I've adjusted the patch to set `=1`, but only if `libei` is enabled.

Maybe we should also warn if Wayland is being used, but not libei? I'm not sure this approach covers _all_ the bases.

shymega:
> > It would also be good to have the check in the CLI interfaces too. Not everyone uses the GUI.
> 
> Did I understand correctly. You mean adding a warning about "wayland support" to stderr? (and input-leap log?)

I mean for when people use `input-leapc` and `input-leaps` directly. The GUI is one thing, but this would be difficult to implement, as we currently use the Qt toolkit to detect Wayland.

CMakeLists.txt
```diff
@@ -172,6 +173,12 @@ if (UNIX)
             install(FILES "dist/flatpak/input-leap-flatpak"
                     DESTINATION bin
                     PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+
+	    if (INPUTLEAP_WARN_ON_WAYLAND)
+                add_definitions(-DINPUTLEAP_WARN_ON_WAYLAND=1)
```
p12tic, line 178: Band indentation.

src/client/input-leapc.cpp
```diff
@@ -44,6 +44,22 @@ int client_main(int argc, char** argv)
     Log log;
     EventQueue events;
 
+    // TODO: Remove once Wayland support is stabilised.
+
+    // This block only warns when `libportal` and `libeis` aren't available - as well as if the top-level CMake option is enabled - by default, it is.
```
p12tic, line 49: Please keep lines to 100 chars or less.

p12tic: Also move this check to separate function.

p12tic:
Several trivial comments.

CMakeLists.txt
shymega: Fixed in 875775d8. Also moved the block to a more logical place.

src/client/input-leapc.cpp
shymega: Also did the same for the server. We don't need to do this for the daemon - that's Win32. See: fe1dd285

shymega:

p12tic:
I think we no longer need this PR, because InputLeap supports Wayland completely and the warning that we want to disable has been removed.

shymega:
It doesn't support it _completely_ yet. Wlroots for example. But for the majority, yes. I will close this PR.

PR #1689, opened 2023-09-19
Add warning when running IL GUI from an Git untagged build

shymega:
The reason for this PR is to be able to warn users about running Input Leap from an untagged Git commit. We should not apply this PR until v3.0.0 of Input Leap is released.

Remaining tasks:

* [ ] Add CMake warning during build process.
* [ ] Add CMake scaffolding for determining Git status, and passing as compiler option.
* [ ] Add warning to CLI tools.
* [ ] Add newsfragment.

## Contributor Checklist:

### DIFF
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -23,6 +23,7 @@
 #include "MainWindow.h"
 #include "AppConfig.h"
 #include "SetupWizard.h"
+#include "common/GitChecks.h"
 
 #include <QtCore>
 #include <QtGui>
@@ -110,6 +111,17 @@ int main(int argc, char* argv[])
 	{
 		return 1;
 	}
+#endif
+
+    // Verify if build was built from a Git tag, or from a Git branch/untagged commit.
+    // TODO: Add CMake support.
+    // For now, this `QMessageBox` is disabled.
+#if defined(INPUTLEAP_GIT_UNTAG_WARNING)
+    if (1)
+        QMessageBox::warning(
+            nullptr, "Input Leap",
+            "This build of Input Leap was built from an untagged commit.\n"
+            "We recommend using a Git tag to build. Here be dragons");
 #endif
 
 	int trayAvailable = waitForTray();
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,9 @@ option(INPUTLEAP_USE_EXTERNAL_GTEST "Use external installation of Google Test fr
 option(INPUTLEAP_BUILD_X11 "Build with XWindows support" ON)
 option(INPUTLEAP_BUILD_LIBEI "Build with libei support" OFF)
 option(INPUTLEAP_BUILD_GULRAK_FILESYSTEM "Use internal filesystem library" OFF)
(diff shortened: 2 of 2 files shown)


p12tic:
If users are building InputLeap themselves, they already know what they are doing. I believe an additional warning is not needed.

shymega:
I disagree. The CI will still build from Git, and people do use the CI artifacts.

I am trying to reduce the amount of bug reports where a Git build was used, and there was no proper warning in place.

I think this addition will assist with that. I need to find a way to determine the checkout SHA, but it shouldn't be
too difficult.

Seeing as the warning can be disabled, I don't think it'll be in the way too much.

p12tic:
Sorry for late response.

A proper way to address this would be to have different configurations for building releases and non-releases. For non-releases we should add a warning unconditionally. This way we wouldn't need extra dependency.

shymega:
I'm still in favour of this PR. The warning can be disabled. I want to reduce the burden of unnecessary issues. Once we make a release, it's less of a problem, but this is an opt-out feature. There is no harm in having it.

shymega:
> Sorry for late response.
> 
> A proper way to address this would be to have different configurations for building releases and non-releases. For non-releases we should add a warning unconditionally. This way we wouldn't need extra dependency.

Agreed. Can we achieve this with CMake? I'd use `#if` to check for the release type during build, and for the opt-out `option` defined.

p12tic:
@shymega Sorry for late comment again!

You could add `option(INPUTLEAP_RELEASE_BUILD "This is a official release" OFF)` to root CMakeLists.txt. And then adjust INPUTLEAP_VERSION_STAGE depending on it. Right now it defaults to "release".

Overall I think Version.cmake could have some cleanup, but this is separate matter.

src/gui/src/main.cpp
```diff
@@ -112,6 +113,17 @@ int main(int argc, char* argv[])
 	}
 #endif
 
+    // Verify if build was built from a Git tag, or from a Git branch/untagged commit.
+    // TODO: Add CMake support.
+    // For now, this `QMessageBox` is disabled.
+#if defined(INPUTLEAP_GIT_UNTAG_WARNING)
+    if (1)
```
p12tic, line 120: Maybe it makes sense to add a yellow bar at the top of the GUI window and write message there? QMessageBox requires user interaction and will be very inconvenient.

src/lib/common/GitChecks.cpp
```diff
@@ -0,0 +1,11 @@
+#include "GitChecks.h"
```
p12tic, line 1: File no longer needed.

src/lib/common/CMakeLists.txt
```diff
@@ -34,6 +34,14 @@ endif()
 
 add_library(common STATIC ${sources})
 
+include(FetchContent)
```
p12tic, line 37: Change no longer needed.

src/lib/common/GitChecks.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
```
p12tic, line 1: File no longer needed.

shymega:
I haven't forgotten about this one.

PR #1693, opened 2023-09-23
Add SUSE spport to input-leap.spec.in

p12tic:
This extracts part of https://github.com/input-leap/input-leap/pull/1629 that can go in separately.


shymega (commented):
Looks reasonable to me. Makes sense, too. LGTM.

PR #1694, opened 2023-09-23
Cleanup version printing code

p12tic:
## Contributor Checklist:

* [not needed] I have created a file in the `doc/newsfragments` directory *IF* it is a


shymega (commented):
This looks good. I'm currently working on that Rust branch, and part of the initial work is to define static consts with `common/Version` class, but I've found it a bit weirdly defined with it. I'm hoping we can simplify the version handling code, and remove some legacy code in the future, which will pave more obstacles.

PR #1701, opened 2023-10-07
macOS: fix power management assertion handling

Koston-0xDEADBEEF:
Create a single power management assertion ID for announcing user activity on screen enter, instead of creating a new one each time.


Koston-0xDEADBEEF:
This way `pmset -g assertions` output doesn't show a new assertion for Input Leap every time screen is entered; existing assertion is refreshed instead.

p12tic (approved):

Issue #1699, opened 2023-10-03
wayland client lose xdg portal Remote Desktop access after it blanked its screen


prahal:
### What happened?

Running Wayland on the input-leap client, I get the XDG portal popup to allow RemoteDesktop Access when on input-leap client startup.
But then I also get this XDG portal popup to allow RemoteDesktop after each screen blanking.
I can force screen blanking on gnome-shell with:
`busctl --user call org.gnome.Shell /org/gnome/ScreenSaver org.gnome.ScreenSaver SetActive b true`

The input-leap client log shows that the libei device is removed on each screen blanking. I might have to report this issue to mutter or libei too as I saw no code to handle EI device removal in the input-leap repository.

The biggest issue with this behavior is we cannot move the mouse to the client to "wake it" up from screen blanking.
Thus one cannot close the client LID if it is a laptop connected to an external monitor without dedicated external input devices connected to the laptop.
Also even if the laptop keyboard is accessible this behavior requires to periodically click on the XDG portal popup. (with a screen blanking timeout of 2 minutes it becomes annoying).

### Version

From Git HEAD or commit (specify below)

### Git commit hash (if applicable)

c6f097e91

### If applicable, where did you install Input Leap from?

git

### What OSes are you seeing the problem on? (Check all that apply)

Linux

### If applicable, are you using Wayland or X11?

Wayland

### What OS versions are you using?

Debian trixie 20231003

### Relevant log output

```shell
[2023-10-03T00:31:12] DEBUG2: reading secure socket
[2023-10-03T00:31:12] DEBUG2: reading secure socket
[2023-10-03T00:31:12] DEBUG2: want to read, error=2, attempt=1
[2023-10-03T00:31:12] DEBUG2: msg from server: CALV
[2023-10-03T00:31:12] DEBUG2: writef(CALV)
[2023-10-03T00:31:12] DEBUG2: wrote 4 bytes
[2023-10-03T00:31:12] DEBUG2: writef(CNOP)
[2023-10-03T00:31:12] DEBUG2: wrote 4 bytes
[2023-10-03T00:31:12] DEBUG2: writing secure socket:0x562cd122a260



[2023-10-03T00:31:13] DEBUG: removing device InputLeap client shared virtual absolute pointer
[2023-10-03T00:31:13] NOTE: Logical output size: 3840x1080@0.0
[2023-10-03T00:31:13] DEBUG: removing device InputLeap client virtual pointer
[...]

PR #1705, opened 2023-10-09
ei: try to restore RemoteDesktop sessions when we lose them

whot:
Use the restore_token provided in RemoteDesktop.Start to re-initialize our sessions if we get disconnected. This gives us back the session without any need for interactive user input after e.g. a lock screen.

Fixes #1699 

cc @prahal

Note: filing this as draft because I haven't fully debugged this yet and there may be some memleaks, this is just the first working version.

Requires:
- [x] https://github.com/flatpak/xdg-desktop-portal/pull/1128
- [x] [GNOME/xdg-desktop-portal-gnome!108](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/108)
- [x] https://github.com/flatpak/libportal/pull/128

PR #1706, opened 2023-10-09
ei: inhibit idle while we have a RemoteDesktop session active

whot:
Inhibit the session while we have a RemoteDesktop session active.

Filed as Draft until XXX is resolved.

Related to #1699

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/PortalRemoteDesktop.cpp
+++ b/src/lib/platform/PortalRemoteDesktop.cpp
@@ -140,6 +140,18 @@ void PortalRemoteDesktop::cb_session_started(GObject* object, GAsyncResult* res)
                         create_event_data<int>(fd));
 }
 
+void PortalRemoteDesktop::cb_inhibited(GObject* object, GAsyncResult* res)
+{
+    g_autoptr(GError) error = nullptr;
+    auto id = xdp_portal_session_inhibit_finish (portal_, res, &error);
+    if (!id) {
+        LOG((CLOG_ERR "Failed to inhibit session"));
+    } else {
+        // Save the ID somewhere so we can call xdp_portal_session_uninhibit() later
+
+    }
+}
+
 void PortalRemoteDesktop::cb_init_remote_desktop_session(GObject* object, GAsyncResult* res)
 {
     LOG((CLOG_DEBUG "Session ready"));
@@ -168,6 +180,16 @@ void PortalRemoteDesktop::cb_init_remote_desktop_session(GObject* object, GAsync
                           reinterpret_cast<PortalRemoteDesktop*>(data)->cb_session_started(obj, res);
                       },
                       this);
+
+    xdp_portal_session_inhibit (portal_,
+                                nullptr, // parent
+                                "InputLeap remote control",
+                                XDP_INHIBIT_FLAG_IDLE,
+                                nullptr, // cancellable
+                                  [](GObject *obj, GAsyncResult *res, gpointer data) {
+                                      reinterpret_cast<PortalRemoteDesktop*>(data)->cb_inhibited(obj, res);
+                                  },
+                                this);
 }
 
 gboolean PortalRemoteDesktop::init_remote_desktop_session()
--- a/src/lib/platform/PortalRemoteDesktop.h
+++ b/src/lib/platform/PortalRemoteDesktop.h
@@ -45,6 +45,7 @@ private:
     void cb_init_remote_desktop_session(GObject* object, GAsyncResult* res);
     void cb_session_started(GObject* object, GAsyncResult* res);
     void cb_session_closed(XdpSession* session);
+    void cb_inhibited(GObject* object, GAsyncResult* res);
 
     /// g_signal_connect callback wrapper
     static void cb_session_closed_cb(XdpSession* session, gpointer data)


prahal:
Does this PR prevent the input leap client from blanking?

Mind the issue in #1699 is fixed by https://github.com/flatpak/xdg-desktop-portal/pull/1128 which was merged to xdg-desktop-portal for 1.18.1.

Maybe this PR should be closed?

I would be against preventing blanking the screen. I do not always have input in the input leap client and displays consume a lot of energy.

shymega:
This PR should be left open, and a configuration setting for inhibiting blanking.

whot:
>  Though I did not understand if the reconnect can happen during the screen blanking/lock screen or if one has to get the input leap client to first get out of blanking/screen locking before the code would do the auto reconnect

We lose the session as soon as the screen lock/blank kicks in (doesn't matter if automatic screen lock is enabled in the Privacy settings) and we cannot create a session while the screen is locked. So right now with #1705 we can only restore the session *after* hitting a key on the local keyboard (and manually logging in if locking is enabled). That's with Fedora 39, GNOME 45.

prahal:
@whot I am not sure I  understand what you mean by "we lose the session as soon as the screen lock/blank".
Right now I can screen lock the client or have the screen of the client blank without losing my session.
When locked I can even unblank it and enter the password from the server mouse and keyboard.
This is with input leap https://github.com/input-leap/input-leap/commit/08d43e700b93d39f4f8f7630042d838e106aeba0,
Gnome Shell 45.3 (with the triple buffering patchset)
Mutter 45.3
libei1/libeis 1.2.0
libportal1 0.7.1 with the patch for InputCapture support (but not released with these patches so unlikely to be included in Debian until then)
xdg-desktop-portal-gnome 45.1-3
xdg-desktop-portal 1.18.2-1
All are build as Debian packages.

This might be an unwanted behavior but it works for a few months.


I mean I can even lock the screen of the server and move the mouse from the screen-locked input leap server to the input leap client.

whot:
>  When locked I can even unblank it and enter the password from the server mouse and keyboard.

unless this changed recently this was definitely not the behaviour when I tested - as soon as mutter locks the screen the RemoteDesktop session was closed and inputleap had to create a new session (which wasn't possible until unlocked).

prahal:
I also experienced the behavior you described back in October when I reported this issue.

I believe the libportal side of the RemoteDesktop persistence might have given me the behavior I currently experience which is fine to me but might not be wanted as is (ie without this PR included to enable it from input-leap).

Out of the test code-related patches I have these patches added in the Debian package above 0.7.1:

[0001-remote-Add-API-to-allow-for-persistence-on-RemoteDes.patch.txt](https://github.com/input-leap/input-leap/files/14611361/0001-remote-Add-API-to-allow-for-persistence-on-RemoteDes.patch.txt)
[0002-Move-the-generic-session-declarations-to-a-separate-.patch.txt](https://github.com/input-leap/input-leap/files/14611366/0002-Move-the-generic-session-declarations-to-a-separate-.patch.txt)
[0003-session-implement-a-generic-close-closed-handling.patch.txt](https://github.com/input-leap/input-leap/files/14611367/0003-session-implement-a-generic-close-closed-handling.patch.txt)
[0004-Implement-support-for-the-InputCapture-portal.patch.txt](https://github.com/input-leap/input-leap/files/14611369/0004-Implement-support-for-the-InputCapture-portal.patch.txt)

whot:
fwiw, the remote desktop persistence changes only help with getting the session back (without the permissions dialog) after unlocking, we still lose the client session and have to use a local keyboard to log back in. But once the mutter session is unlocked we can re-connect to it using the persistence tokens.

Do you have screen blank or screen lock configured? If the former but not the latter that'd explain it?

prahal:
@whot I pressed Super+L to lock the client screen from the server keyboard.

I only have auto blanking, not auto-locking otherwise though it should not change the behavior if input still works when I trigger it manually.
I do not lose the client session when blanking and/or locking the screen for a few months.

I believe requesting authorization each time you start the input leap client or server defeats a lot of use cases of synergy/barrier/input-leap. If you can easily use the client input devices to ack the authorization then you probably don't need to use remote device input that much.

The multi-screen policy of Gnome is if there is an external monitor and the laptop lid is closed consider that the user does not want to suspend because the laptop is docked. This is my case, except I do not want to plug an external keyboard and mouse into the dock but use the input-leap server ones. If the laptop lid is closed there is no way to access the laptop keyboard and clickpad.
There are laptop stands that cover the laptop keyboard. If input-leap requires the use of inputs from the laptop, one will have to plug in an external keyboard and mouse ... thus I see no point in unplugging them afterward to use the ones from the input leap server desktop next to it (to have to replug them back each time the screen blank).

To me, input-leap needs a policy setting that authorizes it once and for all until this setting is turned off.
Requiring local input devices to provide the authorization (while a policy can be set via ssh) to allow to use remote inputs instead of local inputs seems weird to me (especially if we cannot remove these local inputs as we might need them at a random time to reauthorize the remote inputs when the screen blanks).
If we could authorize the remotedesktop portal  via ssh that would provide a way around, though that would be cumbersome to do it each time.

I also really want to blank my screen... 50W minimum consumption while a screen is on is a lot, especially since if I stop the screen blanking for the external monitor it will also cut blanking for the internal display of the laptop).

PR #1710, opened 2023-10-17
A few minor fixes to the InputCapture backend

whot:
Some fixes to the barrier handling and connection to the right signal

## Contributor Checklist:


shymega:
Thank you for these changes. However, can we get some changes split out into separate PRs? Sorry to be a bother, but it makes it easier to bisect in future should we need to. I think @p12tic would agree.

shymega:
Actually, it's not, strictly speaking, necessary. I'm just going off previous merges.

I'll merge this now, so we have the latest fixes.

Thanks!

PR #1715, opened 2023-10-24
Use Cmake 3.21 and find_package to Find OpenSSL

sithlord48:
Split into #1713  and #1804 
- Require CMake 3.21+ (This needs 3.19 for OpenSSL 3.0+ support.)

## Contributor Checklist:

* [] I have NOT created a file in the `doc/newsfragments` directory *IF* it is a

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -42,20 +42,45 @@ jobs:
     steps:
 
       # Preparation steps
-      - if: matrix.os == 'ubuntu:18.04'
-        name: Add repositories with newer git and cmake
+      - name: preinstall depends
         run: |
           apt-get update -y
           # software-properties-common is needed for add-apt-repository
           apt-get install -y software-properties-common gpg wget
           # actions/checkout@v3 wants newer git than what's in default repositories
           add-apt-repository ppa:git-core/ppa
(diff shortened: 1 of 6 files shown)


sithlord48:
> CMake 3.21 is only 2 years old and will not be available on some still supported Linux distributions that we build against.

The Already has to update the ci's cmake for ubuntu 18.04. Now it just does it for 20.04 as well.

shymega:
Adjusted grammar and spelling in the PR description.

Agree with @p12tic, even though CI supports newer distros, we still have some distro packages that don't have the version of CMake specified in this PR.

If we do merge this if changes are made, please can you add context to the commit message justifying the change?

Thank you.

sithlord48:
I updated cmake specificity in this patch because version 3.18+ is needed to correctly find the new library names for OpenSSL 3.0+ and later when we want to improve packaging 3.21 is better at a few key things in that part then 3.18

shymega:
@p12tic What do you think of bumping the CMake minimum version in later releases? Perhaps we could keep this PR as a draft/closed, then reanimate it for later Input Leap major releases? That way, @sithlord48's effort isn't completely lost, and we have a good starting point for the future needs.

PR #1719, opened 2023-10-30
Fix clazy issues

sithlord48:
- Use Qt5 style connections 
    - Old SIGNAL / SLOT macros are slower and more error prone 
 - Use `const &` for current item and a `const list` for input to all ranged for loops to prevent detachment
 - Use `Q_EMIT` in place of `emit` emitting a signal (added a few missing ones also )

### DIFF
--- a/src/gui/src/ZeroconfService.cpp
+++ b/src/gui/src/ZeroconfService.cpp
@@ -84,7 +84,7 @@ ZeroconfService::~ZeroconfService() = default;
 
 void ZeroconfService::serverDetected(const QList<ZeroconfRecord>& list)
 {
-    for (ZeroconfRecord record : list) {
+    for (const ZeroconfRecord& record : list) {
         registerService(false);
         m_pMainWindow->appendLogInfo(tr("zeroconf server detected: %1").arg(
             record.serviceName));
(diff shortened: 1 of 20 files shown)


src/gui/src/ZeroconfService.cpp
```diff
@@ -67,29 +67,24 @@ ZeroconfService::ZeroconfService(MainWindow* mainWindow) :
...
-        this, SLOT(errorHandle(DNSServiceErrorType)));
+    connect(zeroconf_browser_.get(), &ZeroconfBrowser::error, this, &ZeroconfService::errorHandle);
 }
 
 ZeroconfService::~ZeroconfService() = default;
 
 void ZeroconfService::serverDetected(const QList<ZeroconfRecord>& list)
 {
-    for (ZeroconfRecord record : list) {
+    for (const ZeroconfRecord &record : list) {
```
p12tic, line 87: Prefer to place & with type.

src/gui/src/main.cpp
```diff
@@ -54,7 +54,7 @@ bool checkMacAssistiveDevices();
 
 void copy_qsettings(const QSettings &src, QSettings &dst)
 {
-    auto keys = src.allKeys();
+    const auto keys = src.allKeys();
```
p12tic, line 57: There's no need for `const` for non-reference types. I've yet to hit a bug that `const` would have prevented, however it introduces a lot of visual noise. Little benefit with certain costs.

p12tic (commented):
LGTM except two nitpicks. Thanks!

src/gui/src/main.cpp
sithlord48: This is due to a clazy check for detachment on the ranged for. We can't use `qAsConst` with the type so we make it `const` prevents the issue.
You can read it more about it [here](https://github.com/KDE/clazy/blob/master/docs/checks/README-range-loop-detach.md).

src/gui/src/ZeroconfService.cpp
p12tic: Also in other places :-)

p12tic: Fixed.

p12tic (approved):

PR #1726, opened 2023-11-11
Increase the max corner size to 9999

### LINKED ISSUE #159: Corner Size GUI should allow values higher than 99
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/159
> 
> Issue created on: 2018-10-19 by @demonbane
> Issue last updated on: 2021-08-17

The Corner Size control in the screen settings and advanced server settings GUI looks like it's using the default QSpinBox settings which results in a limit of 99. This isn't very practical on high-dpi screens where 99 pixels in the corner is barely noticeable.

These controls should be updated with a higher [maximum](http://doc.qt.io/qt-5/qspinbox.html#maximum-prop) value to allow values above 99 to be entered without having to resort to manually editing a config file.

---

> Commented on: 2021-08-17 by @TheDiscordian

I just ran into this. I have a multi-monitor setup that wasn't working (Linux <-> Linux Monitor <-> MacBook), when I moved the mouse left of the MacBook, if it was too high or too low, it jumped all the way to the left-most monitor. Simply disabling corners and setting the size to 122 fixed my issue completely, works almost like a normal extended monitor now. A lot of people seem to run into multi-monitor issues, and I feel like simply fixing the UI here would go a long way.

sithlord48:
Fixes #159
 - Increases the max corner size to 9999.

## Contributor Checklist:

* [X] I have created a file in the `doc/newsfragments` directory *IF* it is a

### DIFF
--- a/doc/newsfragments/159-fix-corner-size-value.bugfix
+++ b/doc/newsfragments/159-fix-corner-size-value.bugfix
@@ -1,0 +1,4 @@
+Solve Issue 159:
+  Corner Size GUI should allow higher value than 99(https://github.com/input-leap/input-leap/issues/159)
+
+Changes the maximum corner size to 9999
--- a/src/gui/src/ServerConfigDialogBase.ui
+++ b/src/gui/src/ServerConfigDialogBase.ui
@@ -659,7 +659,11 @@ Double click on a screen to edit its settings.</string>
              </widget>
             </item>
             <item>
-             <widget class="QSpinBox" name="m_pSpinBoxSwitchCornerSize"/>
+             <widget class="QSpinBox" name="m_pSpinBoxSwitchCornerSize">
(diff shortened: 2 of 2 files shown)


shymega:
What's the rationale for '9999' being the new value? Seems awfully high. I'd also want to check the C++ side of things can handle such a high integer.

sithlord48:
> What's the rationale for '9999' being the new value? Seems awfully high. I'd also want to check the C++ side of things can handle such a high integer.

This is backed by an int so 9999 should ok int is at least 2 bytes. i picked 9999 because there is enough room for it no other reason.

shymega:
Fixed commit message, renamed newsfragment to be more descriptive, now merging.

PR #1794, opened 2023-12-02
fix(gui/$DISPLAY): Add guard to X11/XWayland `$DISPLAY` check

### LINKED ISSUE #1792: GUI cannot run in native Wayland-only session
### What happened?

GUI cannot run in native Wayland-only session

* run in `gnome-shell --no-x11` session or `DISPLAY= input-leap`
* it says `The InputLeap GUI requires the DISPLAY environment variable to be set. Quitting...`

### Version

3e68145

from source

Linux

Wayland

Arch Linux, GNOME 45.1

### Relevant log output

_No response_

### Any other information

I've locally removed the code checking for Xwayland, and it runs in wayland-only just fine in both client and server, so IIUC there's no need for such checking nowadays.

shymega:
As reported in #1792, a bug has been noticed where Input Leap GUI
doesn't recognise 'libportal' support as being present on the system,
and thus refuses to start.

This is a compile-time guard, so it's not checked at run-time.

In future, I will refactor this check to be optional (enabled by
default, adjustable by CMake option), and to simply probe libportal
and/or X11 safely, to ascertain whenever it needs the `$DISPLAY`
variable. It is not a perfect fix, and we should look to refactor.

Fixes #1792.
Closes #1800 - this PR and #1800 are mutually exclusive, and only one is
needed at this point. Both address #1792, but I've approached the
problem from two different perspectives.

## Contributor Checklist:

### DIFF
--- a/src/gui/src/DisplayIsValid.cpp
+++ b/src/gui/src/DisplayIsValid.cpp
@@ -1,4 +1,4 @@
-#ifdef WINAPI_XWINDOWS
+#if (defined(WINAPI_XWINDOWS) && (!defined(HAVE_LIBPORTAL_INPUTCAPTURE) || !defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)))
 
 #include "DisplayIsValid.h"
 #include <X11/Xlib.h>
--- a/src/gui/src/DisplayIsValid.h
+++ b/src/gui/src/DisplayIsValid.h
@@ -1,5 +1,5 @@
 #pragma once
 
-#ifdef WINAPI_XWINDOWS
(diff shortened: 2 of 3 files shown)


shymega:
I haven't added a newsfragment yet. This needs to be done before merging. I'm as of yet not come up with a suitable fragment.

sithlord48 (approved):

sithlord48 (approved):

darkblaze69:
Checked this patch, and it still can't run in wayland in contrast to #1800:

`The InputLeap GUI requires the DISPLAY environment variable to be set. Quitting...`

shymega:
OK, best to merge #1800 then. @sithlord48 Can you review #1800 please? This PR isn't working, so #1800 is fallback.

sithlord48:
Since this one is the one we are not going to land can we close it ?

shymega:
We can - when the other PR lands, this PR will be closed

PR #1797, opened 2023-12-14
Log window fixes (solution for issue 1717)

hunterzero99:
Used a buffer written to screen every 100ms to dramatically lower CPU use of log window.  
Limit log window to 10,000 lines to prevent RAM use from constantly increasing. 

Proposed fix for https://github.com/input-leap/input-leap/issues/1717

### DIFF
--- a/src/gui/src/LogWindow.cpp
+++ b/src/gui/src/LogWindow.cpp
@@ -66,12 +66,6 @@ void LogWindow::appendError(const QString& text)
 void LogWindow::appendRaw(const QString& text)
 {
     buffer += text + '\n';
-
-    // Truncate the buffer if it exceeds the maximum size
-    if (buffer.size() > maxBufferSize)
-    {
-        buffer = buffer.right(maxBufferSize);
-    }
 }
 
 void LogWindow::flushBuffer()
--- a/src/gui/src/LogWindow.cpp
+++ b/src/gui/src/LogWindow.cpp
@@ -19,6 +19,8 @@
 
 #include <QDateTime>
 
+#include <QTimer>
+
 static QString getTimeStamp()
 {
     QDateTime current = QDateTime::currentDateTime();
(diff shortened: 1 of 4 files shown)


src/gui/src/LogWindowBase.ui
```diff
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
+    <width>996</width>
```
sithlord48, line 9: Are these changes intended? Sadly when you edit ui files the designer loves to changes other lines you didn't want to on you.

To avoid editing this file at all in the constructor after `setupUi(this)` you can call  `m_LogOutput->setMaximumBlockSize(10000)`

src/gui/src/LogWindow.cpp
```diff
@@ -32,33 +34,58 @@ LogWindow::LogWindow(QWidget *parent) :
...
-    if (!m_pLogOutput->toPlainText().isEmpty())
-        appendRaw("");
+    if (!buffer.isEmpty())
+        buffer += '\n';
 }
 
 void LogWindow::appendInfo(const QString& text)
 {
-    appendRaw(getTimeStamp() + " INFO: " + text);
+    buffer += getTimeStamp() + " INFO: " + text + '\n';
```
sithlord48, line 53: Dont use concatantation insead do this
create a private for the widget
 `QString m_infoLine = QStringLiteral("%1 INFO: %2 \n");`
use this in place of your string 
`buffer.append(m_infoLine.arg(getTimeStamp(), text));`

using concatanation is very slow each + will cause a new check and resize of the string for the sting on the right side of the + for the string here it happens 3 times then again for the buffer... using append and/or arg subs will make this once for the string and once for the buffer.

Making our line a QStringLiteral will reduse the memory use as the template string is made at build time and stored in the binary using it at runtime costs nothing.

src/gui/src/LogWindow.cpp
```diff
@@ -32,33 +34,58 @@ LogWindow::LogWindow(QWidget *parent) :
     // repeatedly until InputLeap is finished
     setAttribute(Qt::WA_DeleteOnClose, false);
     setupUi(this);
+    
+    // Use a timer to flush the buffer every 100 milliseconds
+    QTimer* timer = new QTimer(this);
+    connect(timer, &QTimer::timeout, this, &LogWindow::flushBuffer);
+    timer->start(100);
```
sithlord48, line 41: Instad of flushing every 100msec perhaps flush when the buffer size is some size. this will prevent us from doing any extra syncs if the log is still small

src/gui/src/LogWindow.cpp
```diff
@@ -32,33 +34,58 @@ LogWindow::LogWindow(QWidget *parent) :
...
+
+void LogWindow::flushBuffer()
+{
+    if (!buffer.isEmpty())
+    {
+        // Insert the new log content from the buffer
+        m_pLogOutput->appendPlainText(buffer);
+        buffer.clear();
+
+        // Scroll to the bottom if it was at the bottom before
```
sithlord48, line 79: Pretty sure that you can set a property in the widget for this behavior

sithlord48 (changes requested):
Nice work!
A few changes could make this preform even better.

src/gui/src/LogWindow.cpp
prahal: @hunterzero99 I considered the review and made https://github.com/prahal/input-leap/tree/wip-log-window-fixes above input-leap/master.

I disabled the scroll-to-bottom code because it only stopped scrolling to the bottom if one clicked the text. Not if one clicked the scrollbar or scrolled with the mouse.
As far as I tested the output in the log window already scrolls with new content if I do not act upon it.

I also made the buffer private.

I did not code much in cpp so I took the liberty to make the "QString m_infoLine = QStringLiteral("%1 INFO: %2 \n");" into private attributes. Maybe there is a cleaner way.

I tested this code.
The only issue I have left is that with input-leap set to DEBUG2 mode and log to file the start of the log is not saved to file.
I believe the log-to-file is made from the LogWindow content and this LogWIndow content being truncated to 10000 lines cut the start of the logs (thus the version and other critical data).

Maybe the log-to-file should be handled by your new flushBuffer instead of proxying through the LogWindow code. That would solve this blocker.

prahal: @sithlord48 about setting a property to get this behavior, did you mean "scroll to the bottom if it was at the bottom before"?
Note this behavior is the default one and this code "to scroll to bottom" introduces bugs (it scrolls to bottom even if we use the mouse to move in the QPlaintTextEdit. So I removed it altogether in https://github.com/prahal/input-leap/commit/2192f7233c4196692d7a90dd24fe68a865e43e0c

@hunterzero99 do you want me to take over with a PR from https://github.com/prahal/input-leap/commit/2192f7233c4196692d7a90dd24fe68a865e43e0c bringing @sithlord48 changes to your code?

I ran input-leap with max debug2 logs and had no hangs for days. My modification looks ready to go to master to me.

hunterzero99: That sounds great to me @prahal, thank you!  I haven't had a chance to build from your changes yet, but they seem good to me at a visual review.

PR #1799, opened 2023-12-19
gui: Do not use generated ui classes as base classes

sithlord48:
Begin Cleaning up the Gui Files. 

This Pr has alot of change but only because we are now using the `ui` files in a way that all other Qt projects do. 
 - Remove the `ui` form being one of the base classes of each Gui Class, In Qt the cpp , h and ui are the "class" normally when ui files are involved.
 - Rename the *Base.ui files to remove Base from the name  The Ui is normally named the same as the class

- All the ts files should have the lines from UI files objects updated to not have "Base" at the end. 
- The Class names in the UI files should have "Base" removed from them 
- Headers will now have a Ui Namespace and a new private member `Ui::Class* ui`
- C++ sources must now access the ui's items via the`ui` namespace
- Any Methods in a header that returned a ui item now has its definition in the cpp file.
- Mainwindow has a new slot to let the SetupWizard toggle server / client mode.


src/gui/src/AboutDialog.h
```diff
@@ -22,17 +23,17 @@
...
 
-class AboutDialog : public QDialog, public Ui::AboutDialogBase
+class AboutDialog : public QDialog
 {
     Q_OBJECT
-
     public:
         AboutDialog(QWidget* parent, const QString& app_name);
+    private:
+        Ui::AboutDialog *ui = nullptr;
```
p12tic, line 37: `*` should positioned near the type, not variable. Please also fix all other occurrences.

p12tic:
The commit messages don't explain what's happening.

E.g. `Gui Cleanup: About Dialog` can mean anything. Something like the following would explain everything much better:

```
gui: Don't derive from generated UI class

This is standard approach in Qt projects
```

Also please split commits to change only one thing. If a commit moves to not derive from generated UI class then it should not contain anything else. Additional changes should be made in subsequent commit.

p12tic:
Additionally, such large PRs are very hard to review. Anything above 500 lines makes sense to split into multiple PRs so that reviewer is not overwhelmed. This allows the PRs to land sooner.

src/gui/src/AboutDialog.h
sithlord48: Is there a code style guide for this project ? , If not lets get one so we don't have to waste time on these style issues. We could also use a git hook to run clang-format or similar to format the code the to our preferred format.

PR #1800, opened 2023-12-19
fix(gui/$DISPLAY): Check for X11 on the Input Leap GUI with Qt

### LINKED ISSUE #1792: GUI cannot run in native Wayland-only session

shymega:
As reported in #1792, a bug has been found where Input Leap GUI doesn't recognise 'libportal' support as being present on the system, and thus refuses to start.

The idea behind this PR only works on the GUI, so we need a long-term solution, but currently, it replaces the X11 check with a call to `QGuiApplication::platformName` instead.

We test for either `xcb` or `wayland`. If X11 is running, we do nothing.

If Wayland is running, I moved the existing Wayland check, to the new `if` block in this commit, and it now displays an Information msgbox to inform the user about current Wayland support.

Fixes #1792.
Closes #1794 - this PR and #1794 are mutually exclusive, and only one is needed at this point. Both address #1792, but I've approached the problem from two different perspectives.

## Contributor Checklist:


darkblaze69:
Can't build:

```
CMake Error at src/gui/CMakeLists.txt:189 (add_executable):
  Cannot find source file:

    src/DisplayIsValid.cpp
```

darkblaze69:
With this patch I finally can run it on native Wayland

sithlord48 (approved):

src/gui/src/main.cpp
```diff
@@ -63,14 +62,18 @@ void copy_qsettings(const QSettings &src, QSettings &dst)
...
-        return 1;
+#if (defined(WINAPI_XWINDOWS) || (defined(HAVE_LIBPORTAL_INPUTCAPTURE) || defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)))
+    const auto platformType = QGuiApplication::platformName();
+
+    if (platformType == "xcb") {
+        // We're running on X11, all good.
+        // Continue running.
+    } else if (platformType == "wayland") {
+        QMessageBox::information(nullptr, "Input Leap",
+                                 "You are using Wayland. Input Leap supports Wayland via `libei`, but not all desktop environment/window managers support our implementation at this time. Therefore, your mileage may vary.");
```
p12tic, line 73: Too long line, please wrap at 100 chars.

p12tic (approved):
LGTM, just a small nitpick

src/gui/src/main.cpp
shymega: @p12tic As someone who is still learning C++, would I need to use string concatenation, or not?

shymega: @p12tic Found a way. Does this latest push resolve this conversation?

shymega:
Rebased against main, now marked as ready for review. When this is merged (@p12tic feel free), it will close the other PR.

shymega:
@p12tic I've wrapped the `#ifdef` as well.

src/gui/src/main.cpp
```diff
@@ -63,14 +62,22 @@ void copy_qsettings(const QSettings &src, QSettings &dst)
...
-    // DISPLAY is bad. Let's check it first and handle it gracefully
-    if (!display_is_valid()) {
-        fprintf(stderr, "The InputLeap GUI requires the DISPLAY environment variable to be set. Quitting...\n");
-        return 1;
+#if (defined(WINAPI_XWINDOWS) || \
+    (defined(HAVE_LIBPORTAL_INPUTCAPTURE) || \
+    defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)))
+    const auto platformType = QGuiApplication::platformName();
+
+    if (platformType == "xcb") {
```
sithlord48, line 70: Please add an else here. There are more platform plugins that Qt can load on linux for these other ones you may want to log or do maybe `QTextstream(stdout) << "The " << platformType << " plugin is not supported";` . Before calling exit

sithlord48 (changes requested):

src/gui/src/main.cpp
shymega: It can load other plugins, but do we need to handle those? Before this PR, all that was handled was Wayland, and X11. I don't think there's any need to handle things specifically, tbh. We could log QPA, but I'm not convinced we need to. Input Leap is noisy in logs enough as it is.

shymega:
Should we do a null pointer check on `const auto platformType`?

sithlord48:
Not needed a plugged must be loaded

src/gui/src/main.cpp
sithlord48: No we don't need to handle them but those plugins are not going to talk to a wayland or X11 server and i think in this case its best for us to just report this platform plugin was loaded and exit.

shymega: It's merged now, but I didn't implement reporting the platform - logging is noisy enough as it is.

PR #1727, opened 2023-11-11
Set initial tab to 0 on ServerConfigDialog window

sithlord48:
Set initial tab to 0 on ServerConfigDialog window

Change the Server Configuration Dialog, so it opens with the Screens
tab, as opposed to the Advanced tab.

## Contributor Checklist:

* [X] I have created a file in the `doc/newsfragments` directory *IF* it is a
      user-visible change (and make sure to read the `README.md` in that directory)

PR #1712, opened 2023-10-20
User Selectable Qt Version at Config time

sithlord48:
## Contributor Checklist:

* [X] I have created a file in the `doc/newsfragments` directory *IF* it is a
      user-visible change (and make sure to read the `README.md` in that directory) 

Fixes: #1395

Allow the user to select the Qt Version if both Qt5 and Qt6 are installed.
 - Set `QT_DEFAULT_MAJOR_VERSION` to 6 to build with Qt6
 - Set `QT_DEFAULT_MAJOR_VERSION` to 5 to build with Qt5 ( Default if not set)
 
Few Changes to the GUI cmakelist
 1. Use `QT_DEFAULT_MAJOR_VERSION` to control the version of Qt Required
          -  Qt5 Builds Require Qt Version 5.9+
          -  Qt6 Builds Require Qt Version 6.2+
 2. Use `QT_DEFAULT_MAJOR_VERSION` in place of `Qt#_Found` for tool guards

PR #1801, opened 2023-12-20
Future

sithlord48:
# DO NOT LAND
A place to test Just a bunch of commits waiting to be landed with eachother

Contains: 
 - #1799 
 - #1712 
 - #1727 
 - #1726 
 - #1719 
 - #1804 
 - #1807 
 - Small "fix" to glue them better (what will happen as they land anyway)
 - Put all the GH jobs into one file


shymega:
I kinda feel like this PR should be on your fork, @sithlord48. Putting it here might make newcomers think it's part of Input Leap when it's not. It also uses our GH Actions minutes, and our organisation isn't on a paid plan. Do you think you could keep it on your fork? Just a thought...

sithlord48:
Will do

PR #1803, opened 2023-12-20
do not require gtest dependency if building without tests

drt1245:
Remove dependency on gtest if tests are disabled (INPUTLEAP_BUILD_TESTS set to OFF)

## Contributor Checklist:


shymega:
Good catch. One more thing before I merge - a newsfragment. The suffix `.bugfix` will be OK.

Thanks.

sithlord48 (approved):
+1

shymega:
Decided to merge without newsfragment.

PR #1804, opened 2023-12-22
Require Cmake 3.21

sithlord48:
Split from #1715 
Bump The required cmake version to 3.21

This is wanted to avoid a possible issue linking OpenSSL 3.0+ in some cases.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -48,6 +48,34 @@ jobs:
     steps:
 
       # Preparation steps
+      - name: preinstall depends
+        run: |
+          apt-get update -y
+          # software-properties-common is needed for add-apt-repository
+          apt-get install -y software-properties-common gpg wget
+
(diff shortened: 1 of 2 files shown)


p12tic:
Do we really need hard dependency on newer CMake? If there are environments which require new cmake for certain functionality, we can throw a warning.

sithlord48:
In general cmake 3.21 is the minimum that would be recommended. For us currently the main benefit is a bug fix for using openSSL < 3.x on windows. However there are other benefits when we start to make packages with cmake.

shymega:
I would rather this PR be rebased against master, not a merge commit. Please undo the merge commit, and rebase against master.

Thank you.

shymega:

shymega (approved):
LGTM, we could do with bumping CMake for sure.

And thanks for sorting out the CI, too!

PR #1805, opened 2023-12-22
Resolve missing Input Leap XDG desktop icon

### LINKED ISSUE #1802: Incorrect app id matching for .desktop icon in Wayland
### What happened?

In Wayland session the app shows generic icon. To correctly display the icon in Wayland session desktop file should provide correct app id. IUUC it should be named like com.github.input-leap.desktop or contain "StartupWMClass=com.github.input-leap" in "Desktop entry". Not sure what's the correct way to make it proper.

Before:

Screenshot from 2023-12-21 01-34-23

After:

Screenshot from 2023-12-21 01-39-35

### Version

3e681454

self built install with patches to allow to run in native Wayland.

Linux

Wayland

Arch Linux, GNOME 45

### Relevant log output

_No response_

### Any other information

_No response_

shymega:
Also, simplify CMake conditionals, legacy code.

Should fix #1802.

## Contributor Checklist:

### DIFF
--- a/res/input-leap.desktop
+++ b/res/input-leap.desktop
@@ -5,5 +5,6 @@ Comment=Keyboard and mouse sharing solution
 Exec=input-leap
 Icon=input-leap
 Terminal=false
-Categories=Utility;RemoteAccess;
+Categories=Utility;Network;
 Keywords=keyboard;mouse;sharing;network;share;
+StartupWMClass=io.github.input_leap.InputLeap
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,11 +377,7 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 
     configure_files (${CMAKE_CURRENT_SOURCE_DIR}/dist/rpm ${CMAKE_BINARY_DIR}/rpm)
     install(FILES res/input-leap.svg DESTINATION share/icons/hicolor/scalable/apps)
-    if("${VERSION_MAJOR}" STREQUAL "2")
-        install(FILES res/input-leap2.desktop DESTINATION share/applications)
-    else()
-        install(FILES res/input-leap.desktop DESTINATION share/applications)
-    endif()
+    install(FILES res/input-leap.desktop DESTINATION share/applications)
 endif()
 
 else()
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -75,8 +75,8 @@ int main(int argc, char* argv[])
     /* Workaround for QTBUG-40332 - "High ping when QNetworkAccessManager is instantiated" */
     ::setenv ("QT_BEARER_POLL_TIMEOUT", "-1", 1);
 #endif
-    QCoreApplication::setOrganizationName("InputLeap");
-	QCoreApplication::setOrganizationDomain("github.com");
+    QCoreApplication::setOrganizationName("input_leap");
+	QCoreApplication::setOrganizationDomain("github.io");
     QCoreApplication::setApplicationName("InputLeap");
 
 	QInputLeapApplication app(argc, argv);


res/input-leap.desktop
```diff
@@ -3,7 +3,8 @@ Type=Application
 Name=Input Leap
 Comment=Keyboard and mouse sharing solution
 Exec=input-leap
-Icon=input-leap
+Icon=input-leap.svg
```
sithlord48, line 6: The icon name without svg should work correctly when inputleap is installed.

darkblaze69:
This still doesn't work - just generic icon.

Screenshot from 2023-12-21 01-34-23

In GNOME looking glass it appears as com.github.input-leap. If I specify `StartupWMClass=com.github.input-leap` in desktop file or rename it to `com.github.input-leap.desktop`, it appears correctly.

Screenshot from 2023-12-23 16-40-56

**After renaming to com.github.input-leap.desktop, it appears correctly in looking glass:**

Screenshot from 2023-12-23 16-46-31

sithlord48:
Can you try using the name `io.github.input_leap.InputLeap` this is what our appstream name would be. Have you installed inputleap ?

darkblaze69:
Yes, I installed it with this patch, StartupWMClass is as in patch, but it's still generic icon.

```
grep -ri leap /usr/share/applications/
/usr/share/applications/input-leap.desktop:Name=Input Leap
/usr/share/applications/input-leap.desktop:Exec=input-leap
/usr/share/applications/input-leap.desktop:Icon=input-leap.svg
/usr/share/applications/input-leap.desktop:StartupWMClass=io.github.input_leap.InputLeap
```

Also doesn't work if I rename this file to `io.github.input_leap.InputLeap.desktop`

Does it show for you the correct icon if run in GNOME wayland by
`DISPLAY= input-leap`

sithlord48:
the icon name should match the desktop name also the icon should be io... also

darkblaze69:
does this patch work for you practically?

sithlord48:
I currently do have not have a wayland session with a DE that has a working portal so i can't really tell you only that in other apps i've done with flatpaks they use the appsteam name and these seam to be also what wayland is looking for along with an icon somewhere in the current theme icon theme or in the generic hicolor theme.

res/input-leap.desktop
shymega: It should, but I wanted to see if adding the extension made a difference.

shymega:
I searched the codebase for 'com.github.*`, and not much appeared except for Flatpak and Snap.

We should be using `io.github.input_leap.InputLeap`. Bizarre.

shymega:
Oh. I see now. It's Qt doing it. This is a simple fix. I'll do it now.

res/input-leap.desktop
shymega: Reverted.

shymega:
@darkblaze69 Hey - would you be able to test the latest (force-pushed) commit? I think I've found the issue.

darkblaze69:
Still not working. But now looking glass shows `io.github.input-leap`

image

shymega:
Apologies, accidentally deleted the branch. Restored. I've asked @sithlord48 on IRC about this, as they're more experienced with Qt than I am. I think I'm on the right track.

sithlord48:
Tldr this patch needs todo the following
1. rename the icon to io.github.input_leap.InputLeap.svg (when installing on linux) 
2. Use this same name for the desktop file (
3. In `/src/gui/src/MainWindow.cpp` In the constructor around line 140 add 
```
#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
    qApp()->setDesktopFileName(QStringLiteral("io.github.input_leap.InputLeap");
#endif
```

This will only work with Qt 5.15 + . And the Desktop file name should match our Rev DNS name from the appstream spec (we should also consider shipping a manifest to be fully appsteam compliant but that is a patch for another day )

sithlord48:
For some reason i was not able to add to this branch so i had to make one on my fork 
https://github.com/input-leap/input-leap/pull/1809 should replace this and be correct @darkblaze69  can you please test this is working correctly for you now.

darkblaze69:
thanks, the new PR looks good in Wayland.

Issue #1789, opened 2023-11-24
Unable to disable clipboard sharing


9001:
### What happened?

I have tried to disable clipboard sharing both through the UI, and by manually editing the config file. Regardless, the checkbox will be re-enabled the next time I visit the settings, and the feature is indeed enabled. The same applies to "max clipboard sharing size", which always returns to 1000000.

I first noticed this after migrating from Barrier 2.4.0 to Input Leap 2.4.0-20230918-git004a54fb (due to upgrading from Fedora 38 to 39); this may have been an issue with Barrier 2.4.0 as well, in which case I just never noticed :)

### Version

From Git HEAD or commit (specify below)

### Git commit hash (if applicable)

3e681454

### If applicable, where did you install Input Leap from?

The issue is observed both when building from source (latest commit as of now), and when installing from the Fedora repositories; https://fedora.pkgs.org/39/fedora-x86_64/input-leap-2.4.0%5E20230918git004a54fb-1.fc39.x86_64.rpm.html

### What OSes are you seeing the problem on? (Check all that apply)

Linux

### If applicable, are you using Wayland or X11?

X11 on the server, Windows on the client

### What OS versions are you using?

Server: Fedora 39
Client: Windows 10 LTSC (1809, 17763.4974)

shymega:
Hi, can you try building after reverting this commit: 906784c06c68e48a439a509536c92ea45d499527

I think that's the cause.

9001:
Thanks for the quick response, and for looking into this :)

I tried reverting the commit, but it doesn't seem to have affected much.

This bug does also occur with the version of InputLeap which is packaged by Fedora 39, which was built from commit 004a54fb , so I think this issue has existed for a bit longer than that.

kkourin:
I think it is because minimum of clipboard size is > 0 (ServerConfigDialogBase.ui)
```
               <property name="minimum">
                <number>1000000</number>
               </property>
```

and this code in ServerConfig.cpp:
```
size_t ServerConfig::setClipboardSharingSize(size_t size) {
    if (size) {
        setClipboardSharing(true);
    } else {
        setClipboardSharing(false);
    }
    using std::swap;
    swap (size, m_ClipboardSharingSize);
    return size;
}
```

So size is always > 0, so clipboard sharing is always set back to true.

kkourin:
[...]

PR #1806, opened 2023-12-22
Fix clipboard sharing settings in GUI

kkourin:
Fixed clipboard sharing toggle in GUI (https://github.com/input-leap/input-leap/issues/1789)

1. Do not turn on clipboard sharing if clipboard sharing is unchecked, even if the clipboard size limit is above 0
2. Save clipboard size limit to generated server configuration file
3. Enable clipboard size spinbox based on clipboard sharing checkbox

## Contributor Checklist:


shymega:
Thanks for this, I'm gonna go ahead and merge. Good catch!

PR #1807, opened 2023-12-23
Remove the (duplicated) Azure CI integration

sithlord48:
Remove the (duplicated) Azure CI integration.

 All of the GH actions are maintained and working. What is the point of having duplicate CI pipelines to maintain that have the same targets? 

Let's make life easier for maintainers, Github Actions, and users.


shymega:
Updated PR description and commit that explains the change, fixed some grammatical errors which made the PR and commit hard to read.

On the subject of the PR itself though, we have GitHub Apps enabled, and @p12tic doesn't want to move away from Azure Pipelines. I cannot influence this.

sithlord48:
Since its currently broken if we plan to keep it it needs to be fixed.

p12tic:
Github builds were more stable lately (last year or so :-) ), so I agree that Azure no longer serves a purpose.

PR #1825, opened 2024-01-31
CI, Mac os: Use the qt-install-action to install Qt

sithlord48:
- Fix the macOS ci by using qt install action to install qt


.github/workflows/builds.yml
```diff
@@ -171,16 +171,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-11, macos-12]
+       include:
+         - os: macos-11
+           qt-version: 5.15
+           qt-major: 5
```
shymega, line 177: I don't like repeating the variables here. Can we use any of the string manipulation functions available in GH Actions to extract the first '5.' or '6.' in `qt-version`?

That way, we can just use one variable.

.github/workflows/builds.yml
```diff
@@ -171,16 +171,34 @@ jobs:
...
 
     env:
       VERBOSE: 1
       TERM: xterm-256color
 
     steps:
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v3
+        with:
+          dir: /Users/runner
```
shymega, line 194: I think I'd rather use a GH Actions context variable here.

.github/workflows/builds.yml
```diff
@@ -190,10 +208,9 @@ jobs:
 
       - name: Setup the build
         run: |
-          . input-leap/macos_environment.sh
```
shymega, line 193: @p12tic Are you happy with this?

shymega (changes requested):
Just a few things.

Also, will this affect the pending PR of #1713?

.github/workflows/builds.yml
sithlord48: In this case there is no reason we don't use it again.
In the future i plan to unify the steps and use this action for all at that point i will use a variable

sithlord48:
> Just a few things.
> 
> Also, will this affect the pending PR of #1713?

This has been cherry picked from #1713 to pull only the qt action since this fixes our Ci build due to os11 taking forever to checkout qt5 from homebrew

.github/workflows/builds.yml
sithlord48: I plan to improve such things when i unify all the build steps

shymega: I think I'd be in favour of improving these things at the beginning.

.github/workflows/builds.yml
shymega: I think I'd be in favour of improving these things at the beginning.

sithlord48: Un needed so remvoed

.github/workflows/builds.yml
shymega: Scratch this. I'm happy with it.

shymega (approved):
Now my requested changes have been actioned, I'm happy with this PR.

LGTM. Merge pending on @p12tic's review, but it would resolve the `macos-11` issues we have now.

shymega:
@p12tic Just need your review now. Commit message seems alright to me, could do with some more description, but not enough of an issue to warrant me to review.

PR #1828, opened 2024-01-31
fix(wayland-warn): Make Wayland warning a compile-time option

shymega:
Tweak the GUI startup warning regarding Wayland to be a compile-time
flag. This is enabled by default, but can be disabled.

I've made this PR a draft, as I'm going to add a newsfragment. I'll do that
shortly, then remove this paragraph with `<del>`.

## Contributor Checklist:

### DIFF
--- a/src/gui/src/main.cpp
+++ b/src/gui/src/main.cpp
@@ -62,9 +62,7 @@ void copy_qsettings(const QSettings &src, QSettings &dst)
 
 int main(int argc, char* argv[])
 {
-#if (defined(WINAPI_XWINDOWS) || \
-    (defined(HAVE_LIBPORTAL_INPUTCAPTURE) || \
-    defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)))
+#ifdef INPUTLEAP_WAYLAND_WARN_USER
     const auto platformType = QGuiApplication::platformName();
 
     if (platformType == "xcb") {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,8 @@ option(INPUTLEAP_USE_EXTERNAL_GTEST "Use external installation of Google Test fr
 option(INPUTLEAP_BUILD_X11 "Build with XWindows support" ON)
 option(INPUTLEAP_BUILD_LIBEI "Build with libei support" OFF)
 option(INPUTLEAP_DEPLOY_FLATPAK_SCRIPT "Deploy Flatpak script [Experimental]" OFF)
+option(INPUTLEAP_WAYLAND_WARN_USER "Warn upon GUI start about Wayland support" ON)
+
 set (CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set (CMAKE_CXX_STANDARD 14)
 set (CMAKE_CXX_EXTENSIONS OFF)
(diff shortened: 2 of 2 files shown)


src/gui/src/main.cpp
```diff
@@ -62,7 +62,8 @@ void copy_qsettings(const QSettings &src, QSettings &dst)
 
 int main(int argc, char* argv[])
 {
-#if (defined(WINAPI_XWINDOWS) || \
+#if (defined(INPUTLEAP_WAYLAND_WARN_USER) && \
+    defined(WINAPI_XWINDOWS) || \
     (defined(HAVE_LIBPORTAL_INPUTCAPTURE) || \
     defined(HAVE_LIBPORTAL_SESSION_CONNECT_TO_EIS)))
     const auto platformType = QGuiApplication::platformName();
```
whot, line 69: Arguably you could move this set of definitions into cmake and just `#define WARN_IF_WAYLAND` (or something), leaving the code to just
```c
#ifdef WARN_IF_WAYLAND
   const auto platformType ...
#endif
```

But the code looks correct, so LGTM!

whot (approved):

src/gui/src/main.cpp
shymega: @whot Interesting suggestion, and I think this would be better to move the definitions to CMakeLists.txt too.

That way, we can use a single `#ifdef` in the headless executables as well, which would link in with your XWayland PR.

Am I right in thinking `check_symbol_exists` sets the `HAVE_*` defines?

whot: The [docs](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html) list this example for checking for `SEEK_SET`:
```
check_symbol_exists(SEEK_SET "stdio.h" HAVE_SEEK_SET)
```
so the variable name is up to you, the `HAVE_FOO` is just a convenient convention.

shymega: I only need to clarify that this would be set at compile-time. So for users who might not _have_ `libei` support in `libportal`, this wouldn't work.

I'm just thinking from a UX POV, checking for the portals at runtime might be more suited.

For now, I'll merge this PR with the CMake modifications, but long term, we should probably check for `libei` at run-time.

whot: fwiw, `PortalRemoteDesktop::cb_session_started()` and `PortalInputCapture::cb_init_input_capture_session()` do this but they hook up via `Event::QUIT`. *could* be hooked up if we have a special event for this but I don't know enough to comment on this.

The main problem with the portal communication is that we need to spin up a glib thread for dbus. And the reason for that is that we use libportal because the dbus portal communication is not always straightforward.

But just to check if the portals are available - this could be done with Qt DBus (however *that* works) because you just need to check if the busname is there, if the respective interface is available and (in the RemoteDesktop case whether it's version 2).

whot: > I only need to clarify that this would be set at compile-time. So for users who might not have libei support in libportal, this wouldn't work.

Correct. Anything in CMake is set at compile-time only.

shymega: I think using DBus to check for buses is a good approach. Would you agree?

That way we can rely on runtime environment checks, which should help with any future bug reports that I can foresee being an issue.

If you could re-review my latest commits, we can merge this one for now...

whot (approved):

src/gui/src/main.cpp
whot: done. And yes, using dbus to check at runtime would be the best approach but I'm not sure I can find the time to learn qt dbus :(

shymega: OK, let's add that to the list.

shymega:
@p12tic Can you review? Do we need a newsfragment?

shymega:

sithlord48:
There is no point in this mr anymore. It's sat so long the dust settled on wayland for most people.  We can put something in the readme Or we need to check at runtime if libei loaded and them report its missing. This flag is just going to be an annoyance to people.

shymega:
Yeah. We can add a runtime check, currently we have compile-time checks for `libei`, so that makes sense. Thank you.

PR #1829, opened 2024-02-01
gui Code cleaning: Action

sithlord48:
Action:
 - Use `snake_case_` style names for private vars
 - Use arg substitution
 - Create a SETTINGS namespace for the settings keys
 - Remove redundant switch case
 - Remove a pointless Assert 
 - Add InputLeap Developers to files copyright

### DIFF
--- a/src/gui/src/Action.h
+++ b/src/gui/src/Action.h
@@ -1,5 +1,6 @@
 /*
  * InputLeap -- mouse and keyboard sharing utility
+ * Copyright (C) 2023-2024 InputLeap Developers
  * Copyright (C) 2012-2016 Symless Ltd.
  * Copyright (C) 2008 Volker Lanz (vl@fidra.de)
  *
--- a/src/gui/src/Action.h
+++ b/src/gui/src/Action.h
@@ -95,6 +95,7 @@ class Action
         static const char* action_type_names_[];
         static const char* switch_direction_names_[];
         static const char* lock_cursor_mode_names_[];
+        static const QString command_template_;
 };
 
 QTextStream& operator<<(QTextStream& outStream, const Action& action);
--- a/src/gui/src/Action.h
+++ b/src/gui/src/Action.h
@@ -43,47 +43,47 @@ class Action
 
     public:
         QString text() const;
-        const KeySequence& keySequence() const { return m_KeySequence; }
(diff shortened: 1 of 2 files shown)


p12tic:
Fixed some nits and dropped `Reorder switch to put the simple cases first` commit. It's easier to understand when all related switch statements have the same order and using the order of enum declaration makes this order consistent.

src/gui/src/Action.h
```diff
@@ -28,6 +29,17 @@ class ActionDialog;
 class QSettings;
 class QTextStream;
 
+namespace SETTINGS {
```
p12tic, line 32: This needs to be camel case, and more descriptive e.g. SettingsKeys

src/gui/src/Action.h
```diff
@@ -42,47 +54,48 @@ class Action
...
-        QString m_SwitchScreenName;
-        int m_SwitchDirection;
-        int m_LockCursorMode;
-        bool m_ActiveOnRelease;
-        bool m_HasScreens;
-
-        static const char* m_ActionTypeNames[];
-        static const char* m_SwitchDirectionNames[];
-        static const char* m_LockCursorModeNames[];
+        KeySequence keySequence_;
```
p12tic, line 86: snake_case_

p12tic:
In the commit messages, start with area like `gui:` or `gui/Action:`. After that, capital letter. In this case all commits should have `gui/Action:` prefix.

src/gui/src/Action.h
sithlord48: Planning to use this namespace for all the settings we have around the various Qt object. 
Using `SETTINGS` as this is not a type you can make just really a way to tag some strings like an enum

src/gui/src/Action.h
sithlord48: All the names here were Camel Case all i did was remove the `m_` and add `_` to the end. Are you saying you want them all to be camel case or the `keySequence_` should be camel case ? 
Can you point me to a style guide I can follow so we stop with such review time wasters.

src/gui/src/Action.h
sithlord48: Done `SettingsKeys` is the namespace

src/gui/src/Action.h
sithlord48: fixed

PR #1846, opened 2024-02-11
Remove macos_environment.sh file no longer to build or for the CI

sithlord48:
- Removes no longer needed `macos_environment.sh` script


shymega:
These are two separate things, in the _same_ PR. Please avoid this. Feedback: split the changes into two PRs.

Thanks.

shymega:
Currently, the commits consist of a Flatpak change, and a macOS change. 

I would rather there were two distinct PRs for the changes, rather than combining.

sithlord48:
Split the flatpak part into #1855

shymega:
@p12tic seems OK with the PR, so I'll go ahead and merge.

9001:
`macos_environment.sh` is still a required file in order to build input-leap for macos, as`./clean_build.sh` tries to source it by default.

I was able to build input-leap on macos after restoring this file, but I also had to disable tests, since gtest was failing to build.

Has the build procedure for macos changed?

sithlord48:
Use cmake or do it from QtCreator. You can check out what the[ CI ](https://github.com/input-leap/input-leap/blob/master/.github/workflows/builds.yml#L246)does it.

9001:
Alright 👍 but that means the wiki should be updated, because that's still recommending the use of clean_build.sh, which is why I ran into this :>

sithlord48:
> Alright 👍 but that means the wiki should be updated, because that's still recommending the use of clean_build.sh, which is why I ran into this :>

Feel free to propose any updates you think the wiki needs.

shymega:
> Alright 👍 but that means the wiki should be updated, because that's still recommending the use of clean_build.sh, which is why I ran into this :>

No, we still use `clean_build.sh`. I've got a fix pushed to my fork, which I'll merge soon. You shouldn't need to use cmake, or QtCreator to build Input Leap. The script is for easy quick builds for devs, but end-users use it too. 

No need to edit the wiki. I've got a fix, and it's a simple one.

PR #1848, opened 2024-02-12
gui: MainWindow, Set MenuRoles Where possible

sithlord48:
Set [menuRoles](https://doc.qt.io/qt-6/qaction.html#MenuRole-enum) for the items where possible, this will act as a hint for window managers to put these menu items in the correct location (i.e mac os puts settings under the menu with the app name)

 - Configure Settings ->    `PreferencesRole`
 - Quit -> `QuitRole`
 - About InputLeap -> `AboutRole`

### DIFF
--- a/src/gui/src/MainWindow.ui
+++ b/src/gui/src/MainWindow.ui
@@ -371,6 +371,9 @@
    <property name="shortcut">
     <string notr="true"/>
    </property>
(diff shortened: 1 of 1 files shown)


shymega (approved):
LGTM. I'm going to merge this w/o other reviews, because it's a simple change, and helps with accessibility.

PR #1854, opened 2024-02-16
ei: fix the scroll events

### LINKED ISSUE #1840: Mouse wheel scroll jumps are huge with Wayland
### What happened?

When I run the IL server on Wayland (and the client on X11 -- although I don't know if this is particularly relevant to this issue -- I don't have Wayland running machine to test as a client) the mouse's scroll wheel events are huge on the client.

That is, when I move the scroll wheel on the mouse just one _notch_ the scrolling jump on the application on the client is huge, as if the wheel were moved much much more than it was.

### Version

c5bb9dca

https://copr.fedorainfracloud.org/coprs/ofourdan/input-leap-ei-enabled/

Linux

Wayland on the server and X11 on the client.

Fedora 39 on the server and Fedora 38 on the client.

### Relevant log output

_No response_

### Any other information

_No response_

whot:
We may get scroll events or scroll discrete events and both may be in fractions of a single logical scroll event. Handle this by mapping 10 pixels to one wheel click (that's what mutter/gtk do too) and otherwise keeping the remainder around in the device for the next scroll event.

Note that we're always sending things through as wheel, so on the client side we only ever call ei_device_scroll_discrete().

In theory, our pointer and keyboard device may be the same ei device, so let's assume we have a user data on all of them and free them during cleanup - that way we don't need complicated checks of which devices are the same device.

Fixes #1840

cc @brianjmurrell, @prahal

---
Note that there's a remaining issue: mutter seems to be sending both normal and discrete scroll events so the scrolling is still twice what it should be. That seems to be a bug in mutter somewhere, not in input-leap.

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -29,6 +29,7 @@
 #include "base/Stopwatch.h"
 #include "base/IEventQueue.h"
 
+#include <cmath>
 #include <cstring>
 #include <cstdlib>
 #include <algorithm>
@@ -36,6 +37,10 @@
 #include <vector>
 
 #include <libei.h>
+
+struct ScrollRemainder {
+    double x, y; // scroll remainder in pixels
+};
 
 namespace inputleap {
 
@@ -85,11 +90,25 @@ EiScreen::~EiScreen()
     events_->set_buffer(nullptr);
     events_->remove_handler(EventType::SYSTEM, events_->getSystemTarget());
 
-    ei_device_unref(ei_pointer_);
-    ei_device_unref(ei_keyboard_);
-    ei_device_unref(ei_abs_);
+    if (ei_pointer_) {
+        free(ei_device_get_user_data(ei_pointer_));
+        ei_device_set_user_data(ei_pointer_, nullptr);
+        ei_device_unref(ei_pointer_);
+    }
+    if (ei_keyboard_) {
+        free(ei_device_get_user_data(ei_keyboard_));
+        ei_device_set_user_data(ei_keyboard_, nullptr);
+        ei_device_unref(ei_keyboard_);
+    }
+    if (ei_abs_) {
+        free(ei_device_get_user_data(ei_abs_));
+        ei_device_set_user_data(ei_abs_, nullptr);
+        ei_device_unref(ei_abs_);
+    }
     ei_seat_unref(ei_seat_);
     for (auto it = ei_devices_.begin(); it != ei_devices_.end(); it++) {
+        free(ei_device_get_user_data(*it));
+        ei_device_set_user_data(*it, nullptr);
         ei_device_unref(*it);
     }
     ei_devices_.clear();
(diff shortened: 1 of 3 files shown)


hunterzero99:
I gave the branch a try, on a Windows client I not get no movement with small mouse wheel scrolls, and with a very agressive scroll I may get a very small amount of movement in certain windows (Google Chrome)

shymega:
> I gave the branch a try, on a Windows client I not get no movement with small mouse wheel scrolls, and with a very agressive scroll I may get a very small amount of movement in certain windows (Google Chrome). It might be a client issue if you're not seeing it on other builds. I'm running an outdated version of barrier on this client due to there being no Windows build without debugging enabled.

Ideally, you should be using the _same codebase_ on both clients. We can't promise Barrier will work with Input Leap mixed forever.

However, the changes in this PR only apply to systems using `libei`.

hunterzero99:
> Ideally, you should be using the _same codebase_ on both clients. We can't promise Barrier will work with Input Leap mixed forever.

Understood, I just ran [this debug build](https://github.com/input-leap/input-leap/actions/runs/7927164889/job/21643093472) from this commit on my Windows client and I get the same result.

whot:
>  on a Windows client I not get no movement with small mouse wheel scrolls, and with a very agressive scroll I may get a very small amount of movement in certain windows (Google Chrome). It might be a client issue if you're not seeing it on other builds

can you run the server with logging please and tell me what it says for scroll events? it should log the deltas it's sending so we can narrow down where the issue is.

src/lib/platform/EiScreen.cpp
```diff
@@ -85,11 +94,25 @@ EiScreen::~EiScreen()
     events_->set_buffer(nullptr);
     events_->remove_handler(EventType::SYSTEM, events_->getSystemTarget());
 
-    ei_device_unref(ei_pointer_);
-    ei_device_unref(ei_keyboard_);
-    ei_device_unref(ei_abs_);
+    if (ei_pointer_) {
+        free(ei_device_get_user_data(ei_pointer_));
+        ei_device_set_user_data(ei_pointer_, NULL);
```
sithlord48, line 99: Any reason to use  `NULL` over `nullptr` here ?

whot: oops, no, just decades of C programming coming through :) good catch, fixed, thanks.

whot:
> on a Windows client I not get no movement with small mouse wheel scrolls, and with a very agressive scroll I may get a very small amount of movement in certain windows (Google Chrome)

Found the bug, switching to draft so this one doesn't get merged accidentally.

whot:
Ready now, I got the units wrong so it worked between ei->ei but not between ei->other.

Note that we still get a duplicate scroll event from mutter so we scroll twice the amount we should but that's a mutter bug.

hunterzero99:
Latest update is working great with a Windows client, thank you for the hard work!

sithlord48 (approved):

prahal:
Confirmed PR working on an IL Wayland server to an IL Wayland client (both Gnome Shell 45.3).

I also confirm the double scroll event on the client from the IL server scroll, that is a page up is 10 scroll up when using the mouse on the Gnome Shell 45.3 IL server while scrolling 5 times from the server to the IL client is one page up.
Is there a bug already open on mutter gnome gitlab for this issue?

To me, this PR is ready to be merged.

shymega:
Merging. I wonder if this affects KWin too, not just Mutter.

whot:
> I wonder if this affects KWin too, not just Mutter.

Mutter has a specific way of event processing, it converts libinput events to clutter events and those clutter events later pass through the main compositor and eventually get sent to the wayland client or, in our case, via libei to us. For historical reasons mutter gets the smooth scroll event and the discrete scroll event from libinput and processes them as two separate events. That's fine with Wayland because the protocol stacks them together as one logical event, depending on the client version. The ei protocol explicitly states either/or so sending both events results in duplicate events on the client side.

KWin *may* have a similar bug if it has a similar internal implementation but either way, this is very much not a libei/input-leap issue.

Thanks heaps for all the merges!

whot:
mutter PR is now filed as [mutter!3637](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/3637)

PR #1858, opened 2024-02-21
Clean up the logging handling

whot:
Wanted to add some log messages for debugging and ended up fixing a whole bunch of annoyances with the log system.

Notably: using `_X_ATTRIBUTE_PRINTF` (stolen from `XFuncproto.h`) and the cmake change means we have `-Wformat` enabled by default now where available which should catch a few errors quickly. I ran into this when I was trying to printf a `std::string` and the compiler didn't tell me.

The last commit simplifies the actual log buffer handling, hopefully to the same result :)

This is on top of #1854 because I didn't want to have to deal with the rebases.
There are some warnings left which are fixed in #1853.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,11 @@ endif()
 
 set (libs)
 
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag (-Wformat HAVE_WFORMAT)
+if (HAVE_WFORMAT)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wformat")
+endif()
 
 if (INPUTLEAP_BUILD_GUI)
     if (QT_DEFAULT_MAJOR_VERSION EQUAL 5)
--- a/src/lib/base/EventQueue.cpp
+++ b/src/lib/base/EventQueue.cpp
@@ -89,7 +89,7 @@ void EventQueue::set_buffer(std::unique_ptr<IEventQueueBuffer> buffer)
     if (m_events.size() != 0) {
         // this can come as a nasty surprise to programmers expecting
         // their events to be raised, only to have them deleted.
-        LOG((CLOG_DEBUG "discarding %d event(s)", m_events.size()));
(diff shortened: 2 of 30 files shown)


whot:
Hooray, even macos is happy now! :)

CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ endif()
 
 set (libs)
 
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag (-Wformat HAVE_WFORMAT)
+if (HAVE_WFORMAT)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wformat")
```
sithlord48, line 64: Generally its frowned upon to add global flags perhaps we can do this for the affected target(s) using 
[target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html)

sithlord48, line 64: Generally its frowned upon to add global defines. Perhaps we can do this for the affected target(s) using 
[target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) ?

p12tic: But what if we want this flag to apply to all targets? Wouldn't adjusting CMAKE_CXX_FLAGS be appropriate then?

p12tic: Alternatively we would need to add a CMake variable like INPUTLEAP_CXX_FLAGS and pass it to all targets. Looks a bit verbose, no? What are other solutions I'm missing?

p12tic: (I will merge the PR. If we decide this needs changing we can do in a subsequent PR)

sithlord48: sure

p12tic: @sithlord48 but please reply to my comments :)

whot: IMO compiler *warning* options should always be enable globally (unless there's a really good reason not to) since they 99% of the time catch bugs.

I agree the other compiler options should probably be per-target (though again, depending suse-cases).

thanks heaps for teh quick merge.

PR #1862, opened 2024-02-22
ei: implement server-side hotkey handling

### LINKED ISSUE #1711: Binding input to one screen does not work on Wayland
### What happened?

Running `/usr/bin/input-leaps -f --no-tray --debug INFO --use-ei --name foobar --disable-client-cert-checking -c /home/ixevix/.config/barrier/synergy_server_dump3.conf --address :24800` on GNOME Wayland session. I'm connecting from Windows 10 VM and to my surprise it seems to work. However the feature that I'm using to lock the input to either the host or VM doesn't seem to work. In my config section: options I have
`keystroke(Control+BackSpace) = lockCursorToScreen(toggle)` and I tried with that and without that and neither works. Noting here that running on X11 it works. Noting here also that I saw some comment that multiple displays don't work on Wayland yet so I had only one screen on the host while testing this.

### Version

edaa634551eb530a4ae6eaf1d31d62a72d70c961

I've built input-leap for wayland including input-leap-git, libportal-inputcapture, xdg-desktop-portal-git, xdg-desktop-portal-gtk-git and xdg-desktop-portal-gnome-git. All packages from AUR on archlinux except libei which is from extra v1.1.0-1.

Linux

Wayland

Archlinux running GNOME 45 and kernel 6.5.7-arch1-1.

### Relevant log output

_No response_

### Any other information

_No response_

whot:
Add a map that contains the key id and a vector of modifier masks + the assigned hotkey id, then register and lookup those hotkeys as we get the events. This is a bit minimum effort but should get us 80% of the way there, further inspiration can be taken from the X implementation of the same thing.

Note that due to how the InputCapture portal works, this code is not triggered unless we're currently capturing input - and that only happens when we're logically on the remote.

No solution currently exists in the ei backend to capture the key on the server side.

Fixes #1711 (partially)

---

A note here: this is the MVP to get Ctrl+Backspace working and it works in my tests here. The X windows implementation is a lot more complex but I'm running out of time so a mostly-working solution is better than none :)

And as mentioned above, this only works to lock the screen on the remote client, not the local server. Maybe the `GlobalShortcuts` portal is an eventual solution for that but for now this will have to do.

---

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -157,13 +157,28 @@ void EiScreen::warpCursor(int32_t x, int32_t y)
 
 std::uint32_t EiScreen::registerHotKey(KeyID key, KeyModifierMask mask)
 {
-    // FIXME
-    return 0;
+    static std::uint32_t next_id;
+    std::uint32_t id = std::min(++next_id, 1u);
+
+    // Bug: id rollover means duplicate hotkey ids. Oh well.
+
+    auto set = hotkeys_.find(key);
+    if (set == hotkeys_.end()) {
+        hotkeys_.emplace(key, HotKeySet { key });
+        set = hotkeys_.find(key);
+    }
+    set->second.add_item(HotKeyItem(mask, id));
+
+    return id;
 }
 
 void EiScreen::unregisterHotKey(uint32_t id)
 {
-    // FIXME
+    for (auto& set : hotkeys_) {
+        if (set.second.remove_by_id(id)) {
+            break;
+        }
+    }
 }
 
 void EiScreen::fakeInputBegin()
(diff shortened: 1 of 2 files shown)


shymega:
Going to assign a review req to @p12tic.

src/lib/platform/EiScreen.cpp
```diff
@@ -481,6 +517,12 @@ void EiScreen::on_key_event(ei_event* event)
 
     LOG((CLOG_DEBUG1 "event: Key %s keycode=%d keyid=%d mask=0x%x", pressed ? "press" : "release", keycode, keyid, mask));
 
+    if (is_primary_) {
```
sithlord48, line 520: can you do 
```
if (is_primary_ && onHotKey(keyid, pressed mask))
    return;
```

src/lib/platform/EiScreen.cpp
```diff
@@ -733,4 +775,40 @@ IKeyState* EiScreen::getKeyState() const
...
+}
+
+void EiScreen::HotKeySet::add_item(HotKeyItem item)
+{
+    set_.push_back(item);
+}
+
+std::uint32_t EiScreen::HotKeySet::find_by_mask(std::uint32_t mask) const
+{
+    for (auto item = set_.begin(); item != set_.end(); ++item) {
```
sithlord48, line 807: any reason you can't use a `ranged for` here ? (and else where w/ `set_` ? )

src/lib/platform/EiScreen.h
```diff
@@ -141,6 +142,33 @@ class EiScreen : public PlatformScreen {
 #if HAVE_LIBPORTAL_INPUTCAPTURE
     PortalInputCapture* portal_input_capture_;
 #endif
+
+    struct HotKeyItem {
+    public:
+        HotKeyItem(std::uint32_t mask, std::uint32_t id);
+        bool operator<(const HotKeyItem& other) const { return mask_ < other.mask_; };
```
sithlord48, line 149: would it be worth it to add `>` and `==` operators  ?

sithlord48 (changes requested):

p12tic (approved):
LGTM, I will fix some nits.

src/lib/platform/EiScreen.h
p12tic: I think the additional operators are necessary. For example many algorithms are implemented in terms of just `<` anyway. So if we need only `<` then it's fine.

src/lib/platform/EiScreen.cpp
p12tic: Fixed.

src/lib/platform/EiScreen.cpp
p12tic: Fixed.

src/lib/platform/EiScreen.h
whot: just ftr, this is all mostly copy-paste with modifications from the X11 implementation, so some of the decisions are just "whatever was needed over there" :)

PR #1875, opened 2024-02-26
Facilitate selection between MacPorts vs Homebrew if both present

ylluminate:
Current behavior is to ignore Homebrew if MacPorts is present. Some systems have both present and one or the other may be desirable to choose. This simply allows the user to define the preferred macOS package management system via a variable named "LEAP__SYS_PKGMGR" and can be called thusly for Homebrew as preferred with QT5 therefrom (or you could use `macports` otherwise; or simply use the default behavior): 
```
LEAP__SYS_PKGMGR="homebrew" ./clean_build.sh
```

## Contributor Checklist:

### DIFF
--- a/macos_environment.sh
+++ b/macos_environment.sh
@@ -16,8 +16,9 @@ if [ -z "$INPUTLEAP_BUILD_ENV" ]; then
 
     printf "Modifying environment for InputLeap build...\n"
 
-    if command -v port; then
-        printf "Detected Macports\n"
+    # Check if user preference is set for MacPorts
+    if [ "$LEAP__SYS_PKGMGR" = "macports" ]; then
+        printf "Using Macports as specified by user\n"
 
         check_dir_exists '/opt/local/lib/cmake/Qt5' 'qt5-qtbase port'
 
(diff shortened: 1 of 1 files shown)


macos_environment.sh
```diff
@@ -16,8 +16,9 @@ if [ -z "$INPUTLEAP_BUILD_ENV" ]; then
 
     printf "Modifying environment for InputLeap build...\n"
 
-    if command -v port; then
-        printf "Detected Macports\n"
+    # Check if user preference is set for MacPorts
+    if [ "$LEAP__SYS_PKGMGR" = "macports" ]; then
```
p12tic, line 20: If you check both `LEAP__SYS_PKGMGR` and `command -v port` then this if could handle both cases

p12tic: i.e. if LEAP__SYS_PKGMGR == macports OR command -v port.

p12tic: Also I think it's better to reuse INPUTLEAP_BUILD_MACPORTS and  INPUTLEAP_BUILD_BREW instead of introducing LEAP__SYS_PKGMGR.

sithlord48 (changes requested):
I've removed this file in another PR. 

I suggest you try to build without using this file It should work just fine It does on the CI and on my local system. You may need to tell CMake where to find Qt when you configure if you use the online installer but if you have use macports or homebrew it does not matter as CMake will find the depends either way.

shymega:
This would be a user-visible change. Not to all users, but to people compiling from source, yes. Please add a newsfragment.

PR #1887, opened 2024-03-11
x11: log a warning if we're running against Xwayland

whot:
Check if the XWAYLAND extension is available and if so, print a warning. This extension is available in xwayland 23.1.0 and later.

Two comments on the patch:
- the indentation in this file is all over the place with a 50/50 split between tabs and spaces, I stuck with spaces (because `.editorconfig` says so)
- there are heuristics for checking for xwayland (output names, xinput device names) but the extension is the only guaranteed reliable one. It's relatively new but the majority of systems that switch to wayland will also have a relatively new xwayland so I think it's easier to just stick with the simpler call.

## Contributor Checklist:

### DIFF
--- a/doc/newsfragments/warn-on-xwayland.feature
+++ b/doc/newsfragments/warn-on-xwayland.feature
@@ -1,0 +1,1 @@
+InputLeap now warns when connecting to Xwayland.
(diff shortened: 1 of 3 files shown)


shymega:
I'm kinda wondering if long-term, it would be good to move this method to `lib/platform`. Other than that, LGTM.

p12tic (approved):

PR #1892, opened 2024-03-31
Don't use a hidden file for the config on linux

### LINKED ISSUE #1851: man page gives incorrect path to config
### What happened?

The man page(s) says the path to the config is `$HOME/.local/share/input-leap/.input-leap.conf`. However after this didn't work, and doing an `strace`, it is actually looking for `$HOME/.local/share/InputLeap/.input-leap.conf`

### Version

v2.4.0

_No response_

fedora

Linux

_No response_

fedora 39

### Relevant log output

_No response_

### Any other information

_No response_

sithlord48:
Unify `USR_CONFIG_NAME` and `SYS_CONFIG_NAME` into one new `CONFIG_NAME`. This moves the USER based config from `$HOME/.local/share/input-leap/.input-leap.conf` -> `$HOME/.config/Inputleap/input-leap.conf`

Fixes #1851  by correcting the path in the manual.

### DIFF
--- a/src/lib/inputleap/ServerApp.h
+++ b/src/lib/inputleap/ServerApp.h
@@ -118,11 +118,9 @@ private:
 
 // configuration file name
 #if SYSAPI_WIN32
-#define USR_CONFIG_NAME "input-leap.sgc"
-#define SYS_CONFIG_NAME "input-leap.sgc"
+#define CONFIG_NAME "input-leap.sgc"
 #elif SYSAPI_UNIX
-#define USR_CONFIG_NAME ".input-leap.conf"
-#define SYS_CONFIG_NAME "input-leap.conf"
+#define CONFIG_NAME "input-leap.conf"
 #endif
 
 } // namespace inputleap
--- a/doc/input-leaps.1
+++ b/doc/input-leaps.1
@@ -71,7 +71,7 @@ interfaces using port number 24800.
 If no configuration file pathname is provided then the first of the
 following to load successfully sets the configuration:
 .IP
-.I $HOME/.local/share/input-leap/.input-leap.conf
+.I $HOME/.config/InputLeap/input-leap.conf
 .br
 .I /etc/input-leap.conf
 .SH COPYRIGHT
--- a/src/lib/common/unix/DataDirectories.cpp
+++ b/src/lib/common/unix/DataDirectories.cpp
@@ -65,10 +65,10 @@ static fs::path profile_basedir()
 #if defined(WINAPI_XWINDOWS) || defined(WINAPI_LIBEI)
     // linux/bsd adheres to freedesktop standards
     // https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
-    const char* dir = std::getenv("XDG_DATA_HOME");
+    const char* dir = std::getenv("XDG_CONFIG_HOME");
     if (dir != nullptr)
         return fs::u8path(dir);
-    return unix_home() / ".local/share";
+    return unix_home() / ".config/";
 #else
     // macos has its own standards
     // https://developer.apple.com/library/content/documentation/General/Conceptual/MOSXAppProgrammingGuide/AppRuntime/AppRuntime.html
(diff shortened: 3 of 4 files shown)


shymega (commented):
Wait, why is the config file hidden?

sithlord48:
> Wait, why is the config file hidden?

not sure my config doesn't have that file at all on my server I was just updating based upon the user report

shymega:
Yeah, we should patch that in the codebase, the config file shouldn't be hidden. But we'll need to add a migration flow as well, so that we rename the `.input-leap.conf` - like how we do with the `Debauchee` migration flow.

sithlord48:
I'll look into it and update this MR after .

shymega:
Looks okay, but let's update the PR title/description, and add migration for the old path in the codebase, so users have a hassle-free experience. Or we can do that in a later PR.

cc:/ @p12tic

whot:
fwiw, config should go into `XDG_CONFIG_HOME` (aka `$HOME/.config`, the user-equivalent to `/etc`), you're putting it into `XDG_DATA_DIR` (aka `$HOME/.local/share`, the user-equivalent to `/usr/local/share`).

I strongly recommend against using `XDG_DATA_DIR` for config, it'll be confusing and inconsistent with other apps out there.

sithlord48:
Done should use ~/.config/InputLeap/input-leap.conf on linux

whot:
possibly stupid idea, feel free to ignore: default.conf as fallback, otherwise $hostname.conf so you can store multiple configurations in the same (synced-across machines) config directory.

sithlord48:
No. When I tried the file was created for me with no intervention

src/lib/inputleap/ServerApp.h
```diff
@@ -118,11 +118,9 @@ class ServerApp : public App {
 
 // configuration file name
 #if SYSAPI_WIN32
-#define USR_CONFIG_NAME "input-leap.sgc"
-#define SYS_CONFIG_NAME "input-leap.sgc"
+#define CONFIG_NAME "input-leap.sgc"
```
shymega, line 121: This should be split into a separate PR. Currently the PR is for 'not using hidden files for config on Linux' - but this change affects win32.

src/lib/inputleap/ServerApp.h
```diff
@@ -118,11 +118,9 @@ class ServerApp : public App {
...
 #elif SYSAPI_UNIX
-#define USR_CONFIG_NAME ".input-leap.conf"
-#define SYS_CONFIG_NAME "input-leap.conf"
+#define CONFIG_NAME "input-leap.conf"
```
shymega, line 123: Not sure if we should keep a `SYS` config name...

shymega (changes requested):
Just a couple of nitpicks.

shymega:
> No. When I tried the file was created for me with no intervention
> […](#)
> On Tue, May 14, 2024, 3:04 AM Povilas Kanapickas ***@***.***> wrote: I don't see migration code, is there any?

I also want to check this is migrated OK on Windows.

sithlord48:
> > No. When I tried the file was created for me with no intervention
> > […](#)
> > On Tue, May 14, 2024, 3:04 AM Povilas Kanapickas _**@**_.***> wrote: I don't see migration code, is there any?
> 
> I also want to check this is migrated OK on Windows.

Nothing changed on windows to migrate..

src/lib/inputleap/ServerApp.h
sithlord48: Both names are the same so no i don't think we need to store a `sys` config name

src/lib/inputleap/ServerApp.h
shymega: OK, thanks for the clarification.

src/lib/inputleap/ServerApp.h
shymega: OK, thanks for the clarification.

shymega:
OK, @p12tic approved this PR, so I'll go ahead and merge.

shymega (approved):
LGTM.

p12tic:
@shymega 

FYI I didn't approve the PR and my comment "I don't see migration code, is there any?" hasn't been addressed. Now we have a regression where user config is being lost.

PR #1900, opened 2024-04-17
Update `doc/input-leap.conf.example-advanced` to match top comment

tianon:
This looks like it was a WIP that got committed a long time ago but never finished, so I've filled out the configuration matching the top comment's setup description to the best of my knowledge.

(See also https://github.com/input-leap/input-leap/discussions/1899)

## Contributor Checklist:

### DIFF
--- a/doc/input-leap.conf.example-advanced
+++ b/doc/input-leap.conf.example-advanced
@@ -27,23 +27,17 @@ section: screens
 end
 
 section: links
-	# larry is to the right of moe and curly is above moe
-	moe:
-		right = larry
-		up    = curly
+	desktop1:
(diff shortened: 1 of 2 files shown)


shymega:
This should have a newsfragment - given it _is_ a user-visible change. Can you add one, and squash the two commits once that's done? The PR desc will need updating too. Thanks.

tianon:
Ok, fair!  Sorry, I wasn't sure whether fixing an example was considered a user-visible change to the software. :innocent:

I should've read `doc/newsfragments/README.md`! :smile:

Updated both :bow:

(and happy to fixup/adjust/etc more if necessary!)

PR #1910, opened 2024-06-07
Update clean_build.sh to use xcrun rather than xcode-select

Jacoby6000:
On Sonoma 14.4.1 I was not able to build without making this change because the SDK path was wrong. xcrun is provided by xcode and gives an exact path to the SDK. This should prevent similar issues in the future.

I also updated the gitignore because I noticed a couple files being generated that don't seem like they should be checked in to git.

## Contributor Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ CMakeLists.txt.user
 # Transient in-project-directory dependencies
 /deps/
 /out/build/x64-Debug
+bundle
+compile_commands.json
--- a/clean_build.sh
+++ b/clean_build.sh
@@ -18,7 +18,7 @@ B_CMAKE_FLAGS="-DCMAKE_BUILD_TYPE=${B_BUILD_TYPE} ${B_CMAKE_FLAGS:-}"
 if [ "$(uname)" = "Darwin" ]; then
     # macOS needs a little help, so we source this environment script to fix paths.
     [ -e ./macos_environment.sh ] && . ./macos_environment.sh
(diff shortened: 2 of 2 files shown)


shymega:
Please don't merge master into the PR. Instead, rebase. I've reverted and force-pushed, but I'm not planning to merge this PR. It may break older macOS machines.

I would like to be sure it won't break older macOS machines. I do not own a Mac, so I can't test, but this is quite a significant PR to make. I would rather use a macOS version (-ge?) logic to check which path to use.

Jacoby6000:
> Please don't merge master into the PR. Instead, rebase. 
Sorry, about that. I'll take a note for future PRs.

> I would like to be sure it won't break older macOS machines. I do not own a Mac, so I can't test, but this is quite a significant PR to make. I would rather use a macOS version (-ge?) logic to check which path to use.

If it helps with your confidence, the old synergy repo was using xcrun in 2014 for similar environment setup stuff.
https://github.com/search?q=repo%3Ainput-leap%2Finput-leap+xcrun&type=commits

I can also see if I can get this stuff to build on some older macos versions via docker or something if you think it's necessary.  What is your target for the oldest version of macos to support?

shymega:
If Synergy are using it upstream, it's probably alright. Would like to get this merged before our inaugural release. Can you fix the merge conflicts, push, and ping me?

Jacoby6000:
@shymega I believe the PR is ready now.

shymega:
@Jacoby6000 Thanks - I'm just running CI, and once I've reviewed the PR, it'll be automatically merged.

shymega:
CI passed. Now making a review.

shymega (approved):
LGTM after fixes.

PR #1913, opened 2024-06-08
feat: Add support for an 'uninstall' target to CMake

shymega:
This works with Make and Ninja.

## Contributor Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,6 +352,21 @@ else()
     message (STATUS "NOT configuring the installer")
 endif()
 
+#
+# Uninstall target
+#
+
+# uninstall target
+if(NOT TARGET uninstall)
+  configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
+    IMMEDIATE @ONLY)
+
+  add_custom_target(uninstall
+    COMMAND ${CMAKE_COMMAND} -P ${CMAKE_BINARY_DIR}/cmake_uninstall.cmake)
+endif()
+
 enable_testing()
 
 add_subdirectory (src)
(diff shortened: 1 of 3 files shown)


sithlord48:
Why is this in draft, this seams to be complete.

shymega:
I need to add a newsfragment. Will do that this week.

sithlord48:
> I need to add a newsfragment. Will do that this week.

I really think this newfragment obsession needs to be re-evaluated . I don't think we really need them for every tiny change. 
This being a great example most  people will never use this uninstall target its mostly for devs who do a local `make install`. Honestly think it would better and more discoverable to make a build.md as part of the documents that has all the options and other info needed  for devs to build something like [this](https://github.com/sithlord48/ff7tk/blob/master/docs/build.md) . but funny enough i forgot to mention the uninstall target on that repo.

p12tic:
About uninstall target, I think this should be responsibility of the operating system packaging. I would say we don't need uninstall target because we already provide packaging support for many formats. Realistically developers should either use input-leap without installing, or to install using e.g. Debian packaging. Accordingly, uninstall target for `sudo make uninstall` is not necessary.

sithlord48:
This target is not intended to nor would it be able to (since it need the build config)  replace normal package uninstall any more then an `install` target can replace normal packages. This is very much a developer convenience feature for testing installs before packaging anything up .

Here again the news fragment is pointless busy work. For this feature its more helpful to document the uninstall target in the build documentation. I would go so far to say the whole news fragment this is kinda pointless when you can just look at the changes between any commits in git and see what happened . We require merges to land code in to our master/main branch. Every MR should have a good descriptive title saying what it does. The descriptions can be used to add even more detail.  Then the change logs between releases can be auto generated by github's release action. Package maintainers for distros will use those notes. We now have a system where changes are tracked and noted without making more annoying steps for contributors (something this project needs way more of) .All we need to do is ask people now to update the merge title / description (something we as maintainers can even do for them ) and it has 0 effect on the code base or there MR needing to be rebased its much less a burden. I would love to start a discussion about how to better track changes while not burdening contributors any more then necessary.

shymega:
Going ahead with the merge. This is a feature I deem useful for users/beta-testers.

I've added a newsfragment as well.

Just waiting for CI to pass.

sithlord48:
You need to run the jobs this should pass now.

shymega:
Merged.

PR #1918, opened 2024-06-14
libei: Fixed small mouse movements

marcelopio:
When moving sub low pixels of delta on libei, we should buffer the output so that these small movements are accumulated.

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -695,10 +695,18 @@ void EiScreen::on_motion_event(ei_event* event)
          }
 #endif
     } else {
-        LOG_DEBUG("on_motion_event on secondary at (dx,dy)=(%.2f,%.2f)", dx, dy);
-        send_event(EventType::PRIMARY_SCREEN_MOTION_ON_SECONDARY,
-                   create_event_data<MotionInfo>(MotionInfo{static_cast<std::int32_t>(dx),
-                                                            static_cast<std::int32_t>(dy)}));
+        buffer_dx += dx;
+        buffer_dy += dy;
+        auto pixel_dx = static_cast<std::int32_t>(buffer_dx);
+        auto pixel_dy = static_cast<std::int32_t>(buffer_dy);
+        LOG_DEBUG2("on_motion_event(buffer) on secondary at (dx,dy)=(%0.2f,%0.2f)", buffer_dx, buffer_dy);
+        if (pixel_dx || pixel_dy) {
+            LOG_DEBUG("on_motion_event on secondary at (dx,dy)=(%d,%d)", pixel_dx, pixel_dy);
+            send_event(EventType::PRIMARY_SCREEN_MOTION_ON_SECONDARY,
+                       create_event_data<MotionInfo>(MotionInfo{pixel_dx, pixel_dy}));
+            buffer_dx -= pixel_dx;
+            buffer_dy -= pixel_dy;
+        }
     }
 }
 
--- a/src/lib/platform/EiScreen.h
+++ b/src/lib/platform/EiScreen.h
@@ -158,6 +158,9 @@ private:
     std::int32_t cursor_x_ = 0;
     std::int32_t cursor_y_ = 0;
 
+    double buffer_dx = 0;
+    double buffer_dy = 0;
+
     mutable std::mutex mutex_;
 
     PortalRemoteDesktop* portal_remote_desktop_ = nullptr;


shymega:
@whot Can you review this for me? I think @p12tic is away right now.

src/lib/platform/EiScreen.cpp
```diff
@@ -695,10 +695,18 @@ void EiScreen::on_motion_event(ei_event* event)
...
     } else {
-        LOG_DEBUG("on_motion_event on secondary at (dx,dy)=(%.2f,%.2f)", dx, dy);
+        buffer_dx += dx;
+        buffer_dy += dy;
+        LOG_DEBUG("on_motion_event on secondary at (dx,dy)=(%.2f,%.2f)", buffer_dx, buffer_dy);
         send_event(EventType::PRIMARY_SCREEN_MOTION_ON_SECONDARY,
-                   create_event_data<MotionInfo>(MotionInfo{static_cast<std::int32_t>(dx),
-                                                            static_cast<std::int32_t>(dy)}));
+                   create_event_data<MotionInfo>(MotionInfo{static_cast<std::int32_t>(buffer_dx),
+                                                            static_cast<std::int32_t>(buffer_dy)}));
```
whot, line 703: this looks like the `send_event()` should be inside a check for `buffer >= 1`, otherwise you may send a 0/0 motion event.

src/lib/platform/EiScreen.cpp
```diff
@@ -695,10 +695,18 @@ void EiScreen::on_motion_event(ei_event* event)
...
+        if (std::abs(buffer_dx) >= 1){
+            buffer_dx = 0;
+        }
+        if (std::abs(buffer_dy) >= 1){
+            buffer_dy = 0;
+        }
```
whot, line 709: if you get subpixel motion (which you *will* get from libinput even for larger motion) this is buggy. assuming you get events of 0.3 with every motion event so your buffer goes: 0, 0.3, 0.6, 0.9, 1.2. Now you send a motion event of 1 but you reset the buffer, dropping the remaining 0.2. Which means that you lose 20% of every full-pixel event. It's even worse for deltas of 1.9 per delta where you'd lose 90% of the motion due to the reset.

The right code is something like:
```c
buffer_dx += dx;
int32_t delta_dx = (int32_t)buffer_dx;
send_event(delta_dx);
buffer_dx = buffer_dx - delta_dx;
```

whot (changes requested):
I think the remainder handling here is buggy and would result in choppy motion. Quickly hand-typed fixed but no guarantees of syntax correctness ;)

src/lib/platform/EiScreen.cpp
marcelopio: It was the original behaviour, that is why I didn't change it previously, it should be fixed now

src/lib/platform/EiScreen.cpp
marcelopio: Done!

src/lib/platform/EiScreen.cpp
```diff
@@ -695,10 +695,18 @@ void EiScreen::on_motion_event(ei_event* event)
...
 #endif
     } else {
-        LOG_DEBUG("on_motion_event on secondary at (dx,dy)=(%.2f,%.2f)", dx, dy);
-        send_event(EventType::PRIMARY_SCREEN_MOTION_ON_SECONDARY,
-                   create_event_data<MotionInfo>(MotionInfo{static_cast<std::int32_t>(dx),
-                                                            static_cast<std::int32_t>(dy)}));
+        buffer_dx += dx;
+        buffer_dy += dy;
+        auto pixel_dx = (int32_t) buffer_dx;
+        auto pixel_dy = (int32_t) buffer_dy;
```
whot, line 701:
```suggestion
        auto pixel_dx = static_cast<std::int32_t>(buffer_dx);
        auto pixel_dy = static_cast<std::int32_t>(buffer_dy);
```
I think, my C++ isn't great though.

src/lib/platform/EiScreen.cpp
```diff
@@ -695,10 +695,18 @@ void EiScreen::on_motion_event(ei_event* event)
...
+        LOG_DEBUG2("on_motion_event(buffer) on secondary at (dx,dy)=(%0.2f,%0.2f)", buffer_dx, buffer_dy);
+        if (pixel_dx || pixel_dy){
```
whot, line 703:
```suggestion
        if (pixel_dx || pixel_dy) {
```

whot (commented):
two minor nits but this LGTM now otherwise, thanks

p12tic (approved):

PR #1928, opened 2024-06-27
Changed Server IP shortcut to not conflict with Start/Stop

Trial97:
## Contributor Checklist:

### DIFF
--- a/doc/newsfragments/updated-server-ip-shortcut.bugfix
+++ b/doc/newsfragments/updated-server-ip-shortcut.bugfix
@@ -1,0 +1,1 @@
+Updated Server IP shortcut to not conflict with Start/Stop button.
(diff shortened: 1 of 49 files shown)


shymega:
I do not understand the need for this PR. Why have you put a `&amp;` between 'I' and 'P'?

When it's ready to be merged (if), then the two commits should be squashed.

Trial97:
The QT framework allows you to define shortcuts by adding an `&amp;` before the letter you want to use as a shortcut.
Right now the shortcut for the `Start`/`Stop` button conflicts with the `Server IP` one.
When I use server configuration the shortcut `Alt+S` starts and stops the input leap, but when I use client configuration using the same shortcut only selects the `Server IP` input.
This change replaces that shortcut from  `Server IP` with `Alt+P` (from what I saw it was not used on that dialog) and makes the `Start`/`Stop` shortcut consistent between server and client configuration.
As a bonus, I tried to make the translations use the same shortcut for `Server IP`.

shymega:
Apologies! I'm not involved with Qt on this project, and I've been doing a **lot** of HTML recently. I'll await for @p12tic to review, and then we can get this merged.

shymega:
Merging without review, want to get a release out of the door.

PR #1946, opened 2024-08-12
Version from git tag

sithlord48:
## Contributor Checklist:

 - `cmake/Version.cmake` will now generate  a version based on the git tag used
    - Generated format is `MAJOR`.`MINOR`.`PATCH`.`TWEAK`  Tweak is number of commits since last tag
 - Set the project version info in cmake
 - Replace `INPUTLEAP_VERSION_STAGE` use with `INPUTLEAP_VERSION_TWEAK`
 - Remove File `Build.properties`


gabeklavans:
Is the intention to also update the hardcoded default version values (currently set to 2, 4, 0, 0)? 
If so, it could be annoying to have two sources of version info.
If not, then there is the edge case of someone wanting to just download and build the source without git, and their version would be off. Not sure if that's a rational concern, though.

sithlord48:
> Is the intention to also update the hardcoded default version values (currently set to 2, 4, 0, 0) as the version increases? If so, it could be annoying to have two sources of version info. If not, then there is the edge case of someone wanting to just download and build the source without git, and their version would be off. Not sure if that's a rational concern, though.

The hard coded number is only used as a fallback if there is no git info to read . The ONLY way you should see this is if you download src of a commit outside of git. Normal releases will have this in sync as part of the release procedure so their int git and out of git builds will be the same.  Any other builds should be starting with a git checkout  (i.e all our ci  and all other ci i know of )

Koston-0xDEADBEEF:
> > Is the intention to also update the hardcoded default version values (currently set to 2, 4, 0, 0) as the version increases? If so, it could be annoying to have two sources of version info. If not, then there is the edge case of someone wanting to just download and build the source without git, and their version would be off. Not sure if that's a rational concern, though.
> 
> The hard coded number is only used as a fallback if there is no git info to read . The ONLY way you should see this is if you download src of a commit outside of git. Normal releases will have this in sync as part of the release procedure so their int git and out of git builds will be the same. Any other builds should be starting with a git checkout (i.e all our ci and all other ci i know of )

This still means the hardcoded value will get used by essentially all package managers (macports, pkg, dpkg, rpm etc). I'm not sure if this is something input-leap project should care about, though.

sithlord48:
Well that is there problem from not pulling from git. Also those releases should happen on real release versions so in that case the fallback and tag should match. And yeah idk that this is a problem for input-leap to solve.

p12tic:
Ah, now noticed this PR which I reviewed as part of https://github.com/input-leap/input-leap/pull/1947.

p12tic:
What future improvements I see in the version handling:

 - Currently there are 6 places where version information is hardcoded and must be changed on release. Optimally we would reduce this duplication. However, the duplication doesn't cost me much, maybe 2 minutes of time per release, so complicated solutions would be waste of time.
 - There should be a separate setting for CMake to tell that the current build is the actual tagged release. If the setting is not set, then the code should use `-dev` version prefix and find out the git commit of the code that will be compiled.
 
 This way, the version strings will be something like `3.0.1-release` for release and `3.0.1-dev-abcd1234` for the case people download from Github artifacts or build InputLeap themselves.

Changes that do not go into this direction, including part of this PR, will not be accepted.

While I did not agree with the idea of this PR, it contains some good ideas. I will extract the good ideas into separate PRs with correct attribution and merge them separately.

p12tic:
> There should be a separate setting for CMake to tell that the current build is the actual tagged release. If the setting is not set, then the code should use -dev version prefix and find out the git commit of the code that will be compiled.

Expanding on this - there are actually 3 cases:

 - real release with release artifacts being published on Github
 - intermediate builds which people can download from Github artifacts
 - people building themselves
 
 I think it makes sense to have different version string formats for each, so that in bug reports we understand which artifact person used. The last case is especially problematic, because we don't really know what exactly user did to build the executable and what was their environment. So telling users to submit bugs reproduced using Github artifacts would reduce part of the work needed to solve a bug.

PR #1947, opened 2024-08-12
Generated rc

sithlord48:
## Contributor Checklist:

based on #1946 
  - Use Cmake to generate version info in rc files reducing the need to have as many defines


CMakeLists.txt
```diff
@@ -25,6 +25,12 @@ option(INPUTLEAP_USE_EXTERNAL_GTEST "Use external installation of Google Test fr
 option(INPUTLEAP_BUILD_X11 "Build with XWindows support" ON)
 option(INPUTLEAP_BUILD_LIBEI "Build with libei support" OFF)
 option(INPUTLEAP_DEPLOY_FLATPAK_SCRIPT "Deploy Flatpak script [Experimental]" OFF)
+
+include (cmake/Version.cmake)
```
p12tic, line 29: Please make spacing consistent at least in lines that this PR touches.

p12tic (changes requested):
LGTM, just one nit.

CMakeLists.txt
sithlord48: Fixed in #1946 (that should land first since its this items base

p12tic (changes requested):
Looking through the PR a second time, I don't agree with the approach.

The version information should come two ways:

- If it's release, then hardcoded INPUTLEAP_VERSION_* variables. Probably add a INPUTLEAP_IS_RELEASE flag to select this path.

- If it's not release (default), use whatever git supplies.

Overall most the PR is fine, this is just relatively small policy decision.

sithlord48:
> Looking through the PR a second time, I don't agree with the approach.
> 
> The version information should come two ways:
> 
>     * If it's release, then hardcoded INPUTLEAP_VERSION_* variables. Probably add a INPUTLEAP_IS_RELEASE flag to select this path.
> 
>     * If it's not release (default), use whatever git supplies.
> 
> 
> Overall most the PR is fine, this is just relatively small policy decision.

Why over complicate this with another yet another flag? If its a release then the hard code and git tag will match.

shymega:
I agree here with @sithlord48, FWIW.

p12tic:
I believe rc file generation could be submitted as a separate PR, no? https://github.com/input-leap/input-leap/pull/1946 is not going to be merged as is.

sithlord48:
Sure you do what ever you want with it

PR #1957, opened 2024-09-10
flatpak: Update manifest

ofourdan:
Update the input-leap flatpak manifest so that current input-leap and dependencies (including libei) can build.

## Contributor Checklist:

/cc @whot

### DIFF
--- a/dist/flatpak/python3-attrs.json
+++ b/dist/flatpak/python3-attrs.json
@@ -1,0 +1,14 @@
+{
+    "name": "python3-attrs",
+    "buildsystem": "simple",
+    "build-commands": [
+        "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"attrs\" --no-build-isolation"
+    ],
+    "sources": [
+        {
+            "type": "file",
+            "url": "https://files.pythonhosted.org/packages/6a/21/5b6702a7f963e95456c0de2d495f67bf5fd62840ac655dc451586d23d39a/attrs-24.2.0-py3-none-any.whl",
+            "sha256": "81921eb96de3191c8258c199618104dd27ac608d9366f5e35d011eae1867ede2"
+        }
+    ]
+}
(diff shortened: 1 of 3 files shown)


whot:
LGTM, thanks

sithlord48:
I don't expect this to build as there seams to be nothing in there grabbing Qt for us. Have you built this recipe on a clean system ?  Consider starting with the KDE sdk and base image that will get us Qt 6

ofourdan:
> I don't expect this to build as there seems to be nothing in there grabbing Qt for us. Have you built this recipe on a clean system ? 

It built here with:

```sh
$ flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
$ flatpak-builder --force-clean --user --install-deps-from=flathub --repo=repo --install builddir  dist/flatpak/com.github.input-leap.input-leaps.yml
```

> Consider starting with the KDE sdk and base image that will get us Qt 6

OK, we can do that, in which case we need to add the python3-attrs and python3-jinja2 modules to build libei from sources.

/cc @rohitpid as well.

dist/flatpak/python3-attrs.json
```diff
@@ -0,0 +1,14 @@
...
+        "pip3 install --verbose --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"attrs\" --no-build-isolation"
+    ],
+    "sources": [
+        {
+            "type": "file",
+            "url": "https://files.pythonhosted.org/packages/6a/21/5b6702a7f963e95456c0de2d495f67bf5fd62840ac655dc451586d23d39a/attrs-24.2.0-py3-none-any.whl",
+            "sha256": "81921eb96de3191c8258c199618104dd27ac608d9366f5e35d011eae1867ede2"
+        }
+    ]
+}
```
sithlord48, line 14: Please add an empty new line in this and the other json file

ofourdan: I'd rather not, those are autogenerated files by https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator

ofourdan: As a side note, we should be able to drop the dependency on `python3-attrs` once https://gitlab.freedesktop.org/libinput/libei/-/merge_requests/302 lands in libei.

ofourdan:
Closing in favor of [!1958](https://github.com/input-leap/input-leap/pull/1958)

PR #1967, opened 2024-09-14
Move config template and snap 

sithlord48:
## Contributor Checklist:

 - Move `res/config.h.in` -> `src/lib/config.h.in`
   - Move the configure_file for it from `CMakeLists.txt` -> `src/lib/CMakeLists.txt`
   - Use `@Only` to configure the file. to be consistent with our other cmake template files.
 - Move `snap` folder into `dist`, with the other distribution files.
   - Update `snapcraft.yml`  to use the proper appid `io.github.input_leap.InputLeap`


dist/snap/snapcraft.yaml
```diff
@@ -15,7 +15,7 @@ apps:
       # Fallback to XWayland if running in a Wayland session.
       DISABLE_WAYLAND: 1
     desktop: usr/share/applications/input-leap.desktop
-    common-id: com.github.input-leap.input-leap
+    common-id: io.github.input_leap.InputLeap
```
p12tic, line 18: Why the previous value isn't good? We do not own input_leap.github.io or input_leap.github.com.

p12tic (changes requested):
One comment, good otherwise.

p12tic:
@sithlord48 I have directed you multiple times about the format of commit messages that I prefer. Are you trolling me? :)

dist/snap/snapcraft.yaml
sithlord48: because the previous value is not what appstream has agreed upon

sithlord48: >"Each AppStream component has a unique identifier (the value of the <id/> tag), which is unique across all distributions" 
 our name for flathub has to be `io.github.input_leap.InputLeap`  so for snap it must also

sithlord48:
> @sithlord48 I have directed you multiple times about the format of commit messages that I prefer. Are you trolling me? :)

what is wrong with my commit messages?

p12tic:
> what is wrong with my commit messages?

We aren't using conventional commits, thus `fix:`, `release:`, etc are redundant.

p12tic (approved):

PR #1973, opened 2024-09-18
Fix build on FreeBSD

Koston-0xDEADBEEF:
This fixes GUI build error on FreeBSD/X11:

`src/gui/src/MainWindow.cpp:843:9: error: use of undeclared identifier 'kill'`

I'm not very satisfied with this fix, but I was unable to find any meaningful difference in the build command line for this file on Linux vs. FreeBSD. I don't know how signal.h gets included in the Linux build. Shouldn't be any harm doing it explicitly, anyway.

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -56,6 +56,8 @@
 #include <Windows.h>
 #endif
 
+#include <signal.h>
+
 namespace {
 
 static const QString allFilesFilter(QObject::tr("All files (*.*)"));


sithlord48 (approved):

src/gui/src/MainWindow.h
```diff
@@ -26,7 +26,7 @@
 #include <QSettings>
 #include <QProcess>
 #include <QThread>
-
+#include <signal.h>
```
p12tic, line 29: Shouldn't this be <csignal> in C++ code?

p12tic: Looks good otherwise.

shymega:
Rebased against master.

src/gui/src/MainWindow.h
p12tic: Looking into this again, I don't see any calls to things from `signal.h` in this header. What was the compile error? The include should be added to the file where the functionality is used.

p12tic:
I have extended the commit description so that it's clear why this is not `<csignal>` and what was the error that caused the build failure.

p12tic:
Also I put the include to the C++ source file instead of header.

sithlord48:
If we are gonna nit pick commits it should be `fix:` not `Fix`

p12tic:
Indeed, even my modified commit title wasn't actually explaining anything.

sithlord48:
may i suggest 

    fix: Building on Freebsd
    Missing include caused a build failure

p12tic:
I don't think it makes sense to try to use conventional commits. input-leap does not do automatic release note generation.

The `gui: Add missing signal.h include` explains quite clearly what was missing. `Fix build on FreeBSD` does not, one must open the patch to see what has changed.

p12tic (approved):

shymega:
Well, I don't think we should be too strict. I use conventional commits because it works for me, and it's easier for me to write.

Let's be more open to commit messages. We don't want to drive people away. We also don't need to open the patch, @sithlord48's example would be the commit message - not the patch.

I'm trying to encourage development here, and if we nit-pick commits based on the style, we're going to create feelings of resentment and frustration.

sithlord48:
> I don't think it makes sense to try to use conventional commits. input-leap does not do automatic release note generation.

Github releases Very much do

Koston-0xDEADBEEF:
> I don't think it makes sense to try to use conventional commits. input-leap does not do automatic release note generation.
> 
> The `gui: Add missing signal.h include` explains quite clearly what was missing. `Fix build on FreeBSD` does not, one must open the patch to see what has changed.

That is true. But then, the patch is a self-explanatory single line change, so I opted to explain the reason for the change instead.

PR #1975, opened 2024-09-18
fix: show window and file menu on mac os

sithlord48:
## Contributor Checklist:

* [X] This is a user-visible change and I have NOT created a file in the `doc/newsfragments` directory (and made sure to read the `README.md` in that directory)

Mac os by default has a menu for APPNAME (with some items auto moved here by the os) so making our menus on mac os  a little differently
Before: 
Screenshot 2024-09-18 at 8 13 27 PM
Screenshot 2024-09-18 at 8 23 31 PM

After: 
Screenshot 2024-09-18 at 8 20 21 PM
Screenshot 2024-09-18 at 8 21 44 PM
Screenshot 2024-09-18 at 8 22 13 PM

### DIFF
--- a/src/gui/src/MainWindow.cpp
+++ b/src/gui/src/MainWindow.cpp
@@ -273,8 +273,14 @@ void MainWindow::createTrayIcon()
 
 void MainWindow::retranslateMenuBar()
 {
+#ifndef Q_OS_DARWIN
     main_menu_->setTitle(tr("&InputLeap"));
     m_pMenuHelp->setTitle(tr("&Help"));
+#else
+    m_pMenuHelp->setTitle(tr("&File"));
+    main_menu_->setTitle(tr("&Window"));
+#endif
+
 }
 
 void MainWindow::createMenuBar()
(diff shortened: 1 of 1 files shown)


p12tic (approved):

p12tic:
@sithlord48 input-leap does not use `fix:` or similar prefixes in the commit messages.

shymega:
> @sithlord48 input-leap does not use `fix:` or similar prefixes in the commit messages.

I use that all the time @p12tic. I don't think we should be picky with commit messages.

Any and all help should have a low barrier to entry.

p12tic:
> I use that all the time @p12tic. I don't think we should be picky with commit messages.

I disagree. Commit messages become important once something breaks in two years and one needs to dig through history to understand what was happening. Consider why most large projects _are_ picky with commit messages and will fail a PR if something trivial such as capitalization is wrong.

p12tic:
> Any and all help should have a low barrier to entry.

Agree, that's why I just fixed myself. But I'm sure I mentioned the same thing before.

shymega:
OK. Can I ask what, if I understand correctly, the issue with conventional commit messages? You can parse them easily, generate changelogs, and they concisely state what's being changed.

Perhaps we should discuss this is a GH Discussion thread.

sithlord48:
github will parse a bunch automaticly we should use them when possible.
i used`gui` cause this is really only a gui thing its not really a `fix` or a `feat` not related to how we `build` has nothing todo with the `ci` either.

PR #1980, opened 2024-10-01
build: remove unused curl dependency

sithlord48:
## Contributor Checklist:

Clean up unneeded dependencies

 - Remove unneeded `curl`

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -58,7 +58,6 @@ jobs:
                   clang \
                   git \
                   libavahi-compat-libdnssd-dev \
-                  libcurl4-openssl-dev \
                   libice-dev \
                   libsm-dev \
                   libssl-dev \
@@ -429,7 +428,6 @@ jobs:
               pkgconf \
               git \
               avahi-libdns \
-              curl \
               libsm \
               libxinerama \
               libxrandr \
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,16 +109,6 @@ if (UNIX)
     find_package (Threads REQUIRED)
     list (APPEND libs Threads::Threads)
 
-    # curl is used on both Linux and Mac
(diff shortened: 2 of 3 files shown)


nbolton:
I can't remember for the life of me why the CURL dep was added. I think because at some stage the GUI would download and install Bonjour from Apple, but it was like 15 years ago.

shymega:
I think the commit msg should explain the justification for the change, and also 'depencency' => 'dependency'. Other than that, I have zero objections.

shymega (approved):
LGTM, and good to have one less dep. Thanks!

nbolton:
> I think the commit msg should explain the justification for the change

Agreed.

shymega:
> > I think the commit msg should explain the justification for the change
> 
> Agreed.

Yeah, I merged after the last amendment, but I think, in future, we should justify commit messages. A one-liner is quite vague.

sithlord48:
I don't understand what you guys mean by justification . `Unused dependency` to me is the justification for removing it. What would a proper message for this be so i know for the future ?

nbolton:
> What would a proper message for this be so i know for the future ?

Good question. I was speaking in general terms, and perhaps in this particular case... what more can you say? I'm struggling to think of an example, to be honest.

sithlord48:
> I think justification was the wrong word I used for this. In short, I got hung up on the commit message, and was nit-picking. Perhaps, in future, we should decide on a case-by-case basis - but the main criteria I generally have for my personal standards are:
> 
>     * Context for the change - i.e, why change X, does it have breaking changes, is there any related PRs/issues that will be fixed, or depend on the change?
> 
>     * Imperative commit messages.
> 
>     * Scope of the change.
> 
>     * Correct grammar and spelling.
> 
> 
> Perhaps we should document this on the wiki, if these are a good baseline for contributors.

Yes i would make a contributing document.

PR #1981, opened 2024-10-01
build: remove stale build scripts

sithlord48:
## Contributor Checklist:

* [ ] This change affects end users and I have created a file in the `doc/newsfragments` directory (and made sure to read the `README.md` in that directory)
* [X] This change does not affect end users

Fixes: #1950 

Matching Wiki Update: https://github.com/input-leap/wiki-prs/pull/14

PR #1982, opened 2024-10-01
build: Remove unused installer script

sithlord48:
## Contributor Checklist:

We missed a script in #1981 
 - Remove unused build script `build_installer.bat`


p12tic:
https://github.com/input-leap/input-leap/pull/1985 will be reverted as it breaks my workflow. Thus this cannot be accepted either. Sorry for wasting your time.

p12tic (changes requested):
https://github.com/input-leap/input-leap/pull/1985 will be reverted as it breaks my workflow. Thus this cannot be accepted either. Sorry for wasting your time.

sithlord48:
> #1985 will be reverted as it breaks my workflow. Thus this cannot be accepted either. Sorry for wasting your time.

well you should update and maintain them since it seams your the only one using them. If you don't update them don't revert the change they are broken as is

PR #1987, opened 2024-10-04
Release fixes

p12tic:
Miscellaneous release-related fixes


sithlord48:
There is No reason to release the Qt5 windows build.. Or really even to have a Qt5 windows build at this point.

p12tic:
Several percent users still use Windows 7 and similar cruft. Let them have this last release and then we can remove Qt5 support.

sithlord48:
> Several percent users still use Windows 7 and similar cruft. Let them have this last release and then we can remove Qt5 support.

They can and should use barrier..

p12tic:
I don't see problems if the release contains Qt5 windows though.

PR #1988, opened 2024-10-04
Update Build Post 3.0.0

sithlord48:
## Contributor Checklist:

 - Remove Support for Qt5 and Qt5 builds
   - Remove Ubuntu 20.04  build (no Qt6)
 - Enable LibEi by default.
 - Bump Mac os needed to 10.15 for x86 build 
 - Remove Mac OS universal ci.
 - Added ubuntu 24.04 ci


p12tic:
From the PR the only thing remaining is removal of Universal macOS build. I think that we should instead fix the universal build and remove per-architecture builds. Multi-architecture builds are the standard way to distribute software on macOS and is expected by macOS users.

PR #1999, opened 2024-10-10
github: Build Opensuse Leap instead of Tumbleweed

p12tic:
Tumbleweed has problems installing dependencies lately (most likely due to glibc upgrade and rebuild). The Docker repository does not provide any specific tags, only latest. As a result, Tumbleweed cannot be depended on for CI builds.

Accordingly, CI is switched to a specific version of OpenSuse Leap distribution.


p12tic:
OpenSuse Leap turns out to be too old. Closing PR.

PR #2010, opened 2024-10-10
Remove the gulrak submodule

KoljaFrahm:
Part of https://github.com/input-leap/input-leap/pull/1984
The gulrak submodule is not required anymore if C++ 17 is used. We can just use the std library. With C++ 17, we get a deprecated warning for Qt::MidButton which fails the build on Ubuntu 22.

## Contributor Checklist:

* [x] This change does not affect end users


KoljaFrahm:
I think at least this can be merged without trouble

p12tic:
> I think at least this can be merged without trouble

Agreed. I will rebase it on top of https://github.com/input-leap/input-leap/pull/2013, as the PRs overlap a bit.

PR #2013, opened 2024-10-10
Remove Qt5 support

p12tic:
Extracted from https://github.com/input-leap/input-leap/pull/1988, authored by @sithlord48

## Contributor Checklist:

* [x] This change affects end users and I have created a file in the `doc/newsfragments` directory (and made sure to read the `README.md` in that directory)
* [ ] This change does not affect end users

PR #2033, opened 2024-10-12
README: Update post 3.0 release

p12tic:
A lot of parts in README no longer apply after 3.0 release.

## Contributor Checklist:

* [x] This change does not affect end users


mirh:
I have seen #2013 and #2026 and I don't want to be that guy, but if dropping older Windows is just about the qt version it's not like there isn't a [patched version](https://github.com/crystalidea/qt6windows7) that works there too.

PR #2086, opened 2024-11-09
fix: Seg fault on arch linux when configuring server. Error with server config dialog show event.

### LINKED ISSUE #2067: Seg fault when attempting to access "Configure Server..." dialog on Arch linux
### What happened?

When clicking the `Configure server...` button in the input-leap GUI, the program crashes. Running input-leap from the terminal and repeating the same steps indicates that the crash is due to a segmentation fault.

### Version

v3.0.2

_No response_

The Arch linux AUR input-leap package

Linux

Wayland

Arch Linux and MacOS Sequoia

### Relevant log output

_No response_

### Any other information

I've made sure that all optional dependencies were installed when installing the input-leap package from the AUR in case the crash was due to missing deps but that didn't fix the issue.

Jakjm:
Fixed a segmentation fault that would occur when configuring servers on certain platforms. This solution was proposed by sithlord48 and they used it to fix the issue for deskflow.
This should fix #2067
## Contributor Checklist:

### DIFF
--- a/doc/newsfragments/arch_bug_when_configuring_server.bugfix
+++ b/doc/newsfragments/arch_bug_when_configuring_server.bugfix
@@ -1,0 +1,1 @@
+Fixed a segmentation fault that would occur when configuring servers on certain platforms. This solution was proposed by sithlord48 and they used it to fix the issue for deskflow.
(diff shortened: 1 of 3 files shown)


shymega:
@p12tic Review for this should be expedited.

PR #2107, opened 2024-11-30
Address issues with modifiers and dead keys on Windows

pedronavf:
After getting a dead key from `ToUnicodeEx`, add an additional `VK_SPACE` to the keyboard state so we reset the dead key flag and subsequent calls with modifiers, like shift, return the right result (-1) instead of 1.

This happened because without reseting the dead key status the new one was attempted to be composed with the old one, which failed and the end result was a single unicode codepoint not marked as a dead key.

This opens the door to potentially use the returned unicode from the second call to `ToUnicodeEx` as the key character instead of maintaining the getDeadKey function.

I submitted this change to Synergy and Barrier (https://github.com/debauchee/barrier/pull/1583), but the Barrier PR was never merged.

## Contributor Checklist:

* [X] This change affects end users and I have created a file in the `doc/newsfragments` directory (and made sure to read the `README.md` in that directory)

### DIFF
--- a/src/lib/platform/MSWindowsKeyState.cpp
+++ b/src/lib/platform/MSWindowsKeyState.cpp
@@ -1349,8 +1349,20 @@ MSWindowsKeyState::getIDForKey(inputleap::KeyMap::KeyItem& item,
 	KeyID id = static_cast<KeyID>(unicode[0]);
 
 	switch (n) {
-	case -1:
-		return inputleap::KeyMap::getDeadKey(id);
+	case -1: {
+        // dead key. add an space to the keyboard so we exit
+        // the dead key mode and future calls to this function
+        // with different modifiers are not affected.
+
+        BYTE emptyState[256] = { };
+        n = m_ToUnicodeEx(VK_SPACE, 0, emptyState, unicode,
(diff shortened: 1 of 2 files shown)


shymega:
@p12tic Can I get your review on this please?

sithlord48:
@pedronavf  does this apply to deskflow also ?

pedronavf:
No, that patch landed on synergy on 2022 (https://github.com/deskflow/deskflow/pull/7149). I submitted the same patch to barrier https://github.com/debauchee/barrier/pull/1583 but it never got merged, and then I found out about input-leap and submitted the PR.

p12tic (approved):
Looks great, thanks!

src/lib/platform/MSWindowsKeyState.cpp
```diff
@@ -1350,6 +1350,17 @@ MSWindowsKeyState::getIDForKey(inputleap::KeyMap::KeyItem& item,
 
 	switch (n) {
 	case -1:
```
p12tic, line 1352: I think whole case block should be put into braces to silence initialization warning.

p12tic: Fixed this now.

PR #2133, opened 2024-12-19
Add a reload action to the tray icon menu.

gichan-jang:
Sometimes the clipboard copy function does not work. In this case, when the input-leap is reloaded, the function works normally again. To make reload easier, add a reload action to the tray icon.

After this patch:

reload_screen_shot

## Contributor Checklist:


shymega:
@p12tic Can I get your review on this please?

p12tic (approved):

p12tic:
> can we merge this, then?

Yes, sure. I went to other things before tests passed.

PR #2151, opened 2025-01-13
feat: make appstream data compliant with flathub standards

tulilirockz:
These changes are required so that the flatpak-builder linter doesnt complain about this files. Ill add the screenshots in a bit.

PR #2154, opened 2025-01-14
feat: use reachable appid by default

tulilirockz:
This pull request is a continuation of #2151, and is related to #2152. This renames all references to our app-id into a reachable one by default, without any downstream modifications. This should satisfy the flathub linters and make it so the app id correctly reflects the repository location.

This changes it from: `io.github.input_leap.InputLeap` (which would resolve to https://github.com/input-leap/InputLeap) to `io.github.input_leap.input-leap`, which resolves corretly to https://github.com/input-leap/input-leap.

This might affect end users if they are doing some goofy stuff with the program files, but it should just work for everyone!

I have to make you all aware that I have not tested if this will break a current installation with an upgrade, though. It should work ootb however just fine.

@shymega friendly ping


tulilirockz:
It builds and runs and shows up just fine on the flatpak build. There might be some funkiness still with other ones but this looks amazing to me

tulilirockz:
This PR should make it possible to submit it to flathub directly without any changes now!

The flatpak works amazingly well

tulilirockz:
If anyone wants to see progress on the flatpak: https://github.com/tulilirockz/flathub/blob/input-leap/io.github.input_leap.input-leap.yaml

nbolton:
Forgive my ignorance, but in reverse domain name notation, with the underscore, wouldn't `io.github.input_leap.input-leap` be equivalent to https://github.com/input_leap/input-leap ? (404) -- Or am I missing something?

tulilirockz:
> Forgive my ignorance, but in reverse domain name notation, with the underscore, wouldn't `io.github.input_leap.input-leap` be equivalent to https://github.com/input_leap/input-leap ? (404) -- Or am I missing something?

Nope! Not missing anything. It is weird but the "input_leap" gets expanded to "input-leap" since the domain cannot have a hyphen before the last segment

nbolton:
> "input_leap" gets expanded to "input-leap"

I had an idea that might be the case. Good to know and very interesting! Thanks.

tulilirockz:
I believe the errors I was getting were because of the github outage, thus it was erroring out as if the domain wasn't reachable. This PR might just not be necessary at all.

tulilirockz:
Nevermind! The linter gets mad when you check the app-id is wrong on the flatpak manifest. 
image

sithlord48:
That's my bad. I forgot that last `-`

tulilirockz:
> That's my bad. I forgot that last `-` 
> You will need to also change the name of the manifest file so its the proper app-id

I believe I did that everywhere on this PR - the manifest also is named correctly i think.

But no problem! we all make mistakes and honestly if it werent from this linter id never know that was a problem

dist/flatpak/io.github.input_leap.InputLeap.yml
sithlord48, line 1: This file should be named `io.github.input_leap.input-leap.yml`  to match the app id

tulilirockz:
Oh damn, need to change the flatpak build appid too

tulilirockz:
Nice! Everything passed now

PR #2158, opened 2025-01-20
Fix keymap not correctly loading for the Ei backend

BMagnu:
On some systems, the keymap file provided by Ei required to create the XKB context is not properly seeked to the start.
This will cause reading the file to fail (or, more accurately, just cause 0 bytes to be read), causing the keymap to just default to a US-keymap.
This PR fixes this, causing the correct keymap to be loaded.

## Contributor Checklist:

### DIFF
--- a/doc/newsfragments/fixed_keymap_loading_fail_on_wayland.bugfix
+++ b/doc/newsfragments/fixed_keymap_loading_fail_on_wayland.bugfix
@@ -1,0 +1,1 @@
+Fixed missing keymap configs on wayland as the XKB keymap from Ei was not properly loading. The respective keymap file was not seeked to the start on some systems.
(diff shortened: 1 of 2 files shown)


sithlord48:
Would this apply to [deskflow](https:://github.com/deskflow/deskflow) also ? If so would you be would to make a patch.

BMagnu:
Yes, it appears that deskflow needs the same patch. I will create that momentarily.
As for me, I simply have just one non-QWERTY keyboard layout set on my system.
But since the keymap file was not seeked, the read got 0 bytes, causing the `LOG_NOTE("Failed to create XKB context: %s", strerror(errno));` to fire, which also left the keymap state uninitialized which ends up defaulting to US keymaps.
To see if this applies to you as well, you can probably check the log and see if you also run into the failure to create the XKB context.

p12tic (approved):

PR #2166, opened 2025-01-26
input-capture: Fix disabled call

3l0w:
The options argument was missing so the argument data wasn't the right one and the method cb_disabled was actually called on a GVariant* instead on a PortalInputCapture*

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/PortalInputCapture.h
+++ b/src/lib/platform/PortalInputCapture.h
@@ -46,7 +46,7 @@ private:
     void cb_init_input_capture_session(GObject* object, GAsyncResult *res);
     void cb_set_pointer_barriers(GObject* object, GAsyncResult *res);
     void cb_session_closed(XdpSession *session);
-    void cb_disabled(XdpInputCaptureSession* session);
+    void cb_disabled(XdpInputCaptureSession* session, GVariant* option);
     void cb_activated(XdpInputCaptureSession* session, std::uint32_t activation_id,
                       GVariant* options);
     void cb_deactivated(XdpInputCaptureSession* session, std::uint32_t activation_id,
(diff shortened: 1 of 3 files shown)


shymega:
@whot @p12tic Can I get your eyes on this one please?

sithlord48:
@3l0w  If this and your other PR also apply to [deskflow](https://github.com/deskflow/deskflow) please consider making PR(s).

src/lib/platform/PortalInputCapture.h
```diff
@@ -58,9 +58,9 @@ class PortalInputCapture {
     {
         reinterpret_cast<PortalInputCapture*>(data)->cb_session_closed(session);
     }
-    static void cb_disabled_cb(XdpInputCaptureSession *session, gpointer data)
+    static void cb_disabled_cb(XdpInputCaptureSession *session, GVariant* options, gpointer data)
     {
-        reinterpret_cast<PortalInputCapture*>(data)->cb_disabled(session);
+        reinterpret_cast<PortalInputCapture*>(data)->cb_disabled(session, options);
```
whot, line 63: semi-unrelated to this patch but it looks like the `cb_activated_cb` and `deactivated` needs fixing too? From the [docs](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.InputCapture.html#org-freedesktop-portal-inputcapture-activated) the `activation_id` is inside the `options` array so on a quick glance the current code is similarly buggy.

whot (approved):
whoopsie. this looks correct to me, thanks.

src/lib/platform/PortalInputCapture.h
3l0w: Nope, i check the signature of activated and deactivated is good because libportal gently extract for us `activation_id` from `option`.
[libportal/inputcapture.c#L420](https://github.com/flatpak/libportal/blob/08c86829ad92d417ffd760f91af3ba9e2f7c52b8/libportal/inputcapture.c#L420)

p12tic:
Looks great, thanks!

PR #2167, opened 2025-01-26
ei: reset w h x y when calculating screen size

3l0w:
These values were not reset each time the screen shape was update, so if the screen shrink the values will be invalid and lead to borders not triggered.
Also x and y need to be initialised at they maximum values or else they will be always 0 since only min are called on them in the calculation processs.

## Contributor Checklist:

### DIFF
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -423,7 +423,10 @@ bool EiScreen::isPrimary() const
 
 void EiScreen::update_shape()
 {
-
+    w_ = 1;
+    h_ = 1;
+    x_ = std::numeric_limits<uint32_t>::max();
+    y_ = std::numeric_limits<uint32_t>::max();
     for (auto it = ei_devices_.begin(); it != ei_devices_.end(); it++) {
         auto idx = 0;
         struct ei_region *r;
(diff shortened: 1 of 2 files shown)


shymega:
@whot @p12tic Can I get your eyes on this one please?

src/lib/platform/EiScreen.cpp
```diff
@@ -423,7 +423,10 @@ bool EiScreen::isPrimary() const
 
 void EiScreen::update_shape()
 {
-
+    w_ = 0;
+    h_ = 0;
+    x_ = 0xFFFFFFFF;
```
p12tic, line 428: std::numeric_limits<std::uint32_t>::max()

p12tic:
One nit, otherwise looks good. Thanks!

src/lib/platform/EiScreen.cpp
3l0w: Thanks

3l0w:
Should be good now

src/lib/platform/EiScreen.cpp
```diff
@@ -423,7 +423,10 @@ bool EiScreen::isPrimary() const
 
 void EiScreen::update_shape()
 {
-
+    w_ = 0;
```
sithlord48, line 426: When the class is created w_ and h_ are set to `1` not 0 .

whot (commented):
Agree with @sithlord48 on the `1` vs `0` but otherwise this LGTM, thanks

src/lib/platform/EiScreen.cpp
3l0w: Should be good now !

PR #2200, opened 2025-03-13
fix: move default input-leap.conf file to `$XDG_CONFIG_HOME` instead of `$HOME`

tulilirockz:


p12tic:
We would need code to migrate from old location.

PR #2226, opened 2025-04-19
fix: Split XKB code out of XWindowsUtil to allow building without X11

chewi:
The xorgproto headers are still required for EI, but nothing more. Note that the GUI hasn't needed X11 for a while.

I dropped the comment in the Flatpak file about this not working, but I have assumed that you still want X11 support enabled there.

I have tested this between two X11-less builds running under KDE Plasma.

We probably don't need to shout about this to users. Most will not build from source, and `-DINPUTLEAP_BUILD_X11=OFF` was supposed to work already.

## Contributor Checklist:


sithlord48:
Hello unfortunately this project is [dead](https://github.com/input-leap/input-leap/discussions/2199). If possible could you try this on the upstream project [deskflow](https://github.com/deskflow/deskflow)

chewi:
Huh, no way. :face_with_diagonal_mouth: I know of this project's history and had come back to it after waiting for the Wayland support to settle and an actual release to happen, but I had no idea about this. I know you only reached a final decision recently, but I strongly suggest you mention this at the top of the README and/or archive the repo. I never look at Discussions on GitHub, and judging by the number of new issues that have been filed recently, neither does anyone else. Thank you for you ongoing efforts with the wider project.

As for the change, Deskflow doesn't event try to make X11 optional, but it looks like it could. I'll see what I can do.

chewi:
Err... right. I'm glad you spoke up because I'm also trying to sort out the packaging on Gentoo Linux. I was already half way to shifting everything over to Deskflow. I don't want to pick sides here, so I may end up packaging both, although Deskflow currently lacks the build improvements I previously made to Barrier.

Thanks for picking this up.

nbolton:
> Deskflow currently lacks the build improvements I previously made to Barrier.

Have you considered porting those features upstream?

nbolton:
> keen to accept them

We're keen to review new PRs and tend to merge within 24 hours in most cases.

shymega:
I would rather we stayed on-topic. Deskflow is welcome to use Input Leap's code,  but please stick to the topic at hand - i.e, a PR to Input Leap.

nbolton:
> Deskflow is welcome to use Input Leap's code

No problem, yes, let's stay on topic about this PR. Input Leap is also welcome to pull [changes from upstream](https://github.com/deskflow/deskflow/commits/master/)! :slightly_smiling_face:

p12tic (approved):

Issue #2221, opened 2025-04-05
Failure to build with Qt 6.9


sithlord48:
### What happened?

Input-leap can not be built with Qt 6.9 (See https://github.com/deskflow/deskflow/issues/8448)

### Version

v3.0.0

### If applicable, where did you install Input Leap from?

N/A

### What OSes are you seeing the problem on? (Check all that apply)

Linux

### What OS versions are you using?

N/A Affects all OS using Qt 6.9.0

### Relevant log output

```shell

```

### Any other information

the changes needed to [fix](https://github.com/deskflow/deskflow/pull/8447/files#diff-047afa1e28908d544cec52f5e815825e52ed59c4681e632582f8af45c86a923c) this have all ready been made on Deskflow. I am Reporting here to make Input-leap users aware of the issue.

PR #2231, opened 2025-04-22
fix(gui): Port Deskflow Qt 6.9 fix

shymega:
As raised in #2221, this fix ports the commit from Deskflow that fixes compilation on machines with Qt versions greater than or equal to 6.9.

See: https://github.com/deskflow/deskflow/pull/8447

Fixes: #2221.

## Contributor Checklist:


shymega:
@p12tic I took this fix from Deskflow, but the `static inline` integer in `KeySequence.h` seems to require C++ STD 17+.

Have you got any other thoughts on how to approach this w/o bumping C++ std on Windows?

sid-code:
Why not just cast it to `uint` as per the [release notes](https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.9.0/release-note.md)?

The relevant section, if I understand it correctly:

> 563ed822f86 QString: toward UTF-8 arg() support [3½/4]: port unary arg() to QAnyStringView The QString::arg() overloads have been redesigned. Character-like types (char, char16_t, char8_t, wchar_t (still subject to QTBUG-126054 at the time of writing), char32_t) now always output the character, not its numeric value. In particular, char16_t and char arguments now match the string-ish arg() overload (and therefore don't provide a base argument anymore. A backwards-compatible fix is to cast char, char16_t, and wchar_t arguments to uint. This also fixes the ambiguity errors you may have seen when using a char16_t as fillChar.

I applied [this patch](https://gist.github.com/sid-code/9131300ee943af1ce5958ee40e527b93) to the source and it builds with Qt 6.9.0.

PR #2247, opened 2025-06-01
Fix missing size_t include in XKBUtil.cpp for stricter compilers

jaded0:
The raspberry pi has a stricter compiler, this is needed in order to build on it.


tianon:
I needed this to compile 3.0.3 on Debian (and adding this one patch was sufficient for the build to succeed). :+1:

p12tic (approved):

PR #2288, opened 2025-08-05
fix(wayland): release portal input capture when EI_EVENT_DISCONNECT is encountered, refactoring

### LINKED ISSUE #2287: Cursor stuck off server screen (Wayland)
### What happened?

If you move the mouse back and forth from the IL gnome shell/mutter 48.3 wayland server to the same setup client after quite many attempts (around 2 minutes) the mouse and keyboard get stuck on the IL gnome shell wayland server.
See deatilas in deskflow bug report https://github.com/deskflow/deskflow/issues/8005 and fix in https://github.com/deskflow/deskflow/pull/8503 (I have prepared a port of this PR).

### Version

v3.0.3

b1f2631ee

self-built

Linux

Wayland

Debian Trixie

### Relevant log output

```shell

```

### Any other information

_No response_

prahal:
See deskflow issue #8005.
Closes #2287

## Contributor Checklist:

### DIFF
--- a/doc/newsfragments/fix_server_input_stuck_when_inputcapture_reconnect.bugfix
+++ b/doc/newsfragments/fix_server_input_stuck_when_inputcapture_reconnect.bugfix
@@ -1,0 +1,1 @@
+Fix InputLeap server inputs stuck when EIS reconnect to client after invalid EIS protocol.
--- a/src/lib/platform/EiScreen.cpp
+++ b/src/lib/platform/EiScreen.cpp
@@ -152,7 +152,7 @@ void EiScreen::cleanup_ei()
             ei_device_set_user_data(ei_abs_, nullptr);
             ei_abs_ = ei_device_unref(ei_abs_);
     }
-    ei_seat_unref(ei_seat_);
+    ei_seat_ = ei_seat_unref(ei_seat_);
     for (auto it = ei_devices_.begin(); it != ei_devices_.end(); it++) {
             free(ei_device_get_user_data(*it));
             ei_device_set_user_data(*it, nullptr);
@@ -738,7 +738,6 @@ void EiScreen::handle_portal_session_closed(const Event &event)
 void EiScreen::handle_system_event(const Event& sysevent)
 {
     std::lock_guard<std::mutex> lock(mutex_);
-    bool disconnected = false;
 
     // Only one ei_dispatch per system event, see the comment in
     // EiEventQueueBuffer::addEvent
@@ -786,10 +785,20 @@ void EiScreen::handle_system_event(const Event& sysevent)
                 // We're using libei which emulates the various seat/device remove events
                 // so by the time we get here our EiScreen should be in a neutral state.
                 //
-                // We don't do anything here, we let the portal's Session.Closed signal
-                // handle the rest.
-                LOG_WARN("disconnected from EIS");
-                disconnected = true;
+                // We must release the xdg-portal InputCapture in case it is still active
+                // so that the cursor is usable and not stuck on the InputLeap server.
+                LOG_WARN("disconnected from eis, will afterwards commence attempt to reconnect");
+                if (is_primary_) {
+                    LOG_DEBUG("re-allocating portal input capture connection and releasing active captures");
+                    if (portal_input_capture_) {
+                        if (portal_input_capture_->is_active()) {
+                            portal_input_capture_->release();
+                        }
+                        delete portal_input_capture_;
+                        portal_input_capture_ = new PortalInputCapture(this, this->events_);
+                    }
+                }
+                this->handle_portal_session_closed(sysevent);
                 break;
             case EI_EVENT_DEVICE_PAUSED:
                 LOG_DEBUG("device %s is paused", ei_device_get_name(device));
@@ -843,9 +852,6 @@ void EiScreen::handle_system_event(const Event& sysevent)
         }
         ei_event_unref(event);
     }
-
-    if (disconnected)
-        ei_ = ei_unref(ei_);
 }
 
 void EiScreen::updateButtons()


nbolton:
Does this fix exist upstream?

sithlord48:
@nbolton  yes we have this upstream. 

@prahal This did not ultimately fix that issue while it helped what ultimately fixed https://github.com/deskflow/deskflow/issues/8005 was https://github.com/deskflow/deskflow/pull/8749

shymega:
Closing for now, as this is incomplete, and not marked as a draft.

Thank you @prahal, but it needs a bit more work on it. Maybe if you want to contribute Deskflow's fixes, we can revisit - but it [this PR] should really be marked a as a draft PR.

prahal:
I totally disagree. This PR is complete and the comment from sithlord48 that talks about the other deskflow PR is about another issue (in fact two others issue mixed in a deskflow PR). Please merge. I am using this PR daily and it fixes the issue reported by me...
I agree that this PR might not fix all bugs that produce an unresponsive mouse... And that is not the point of a PR. It should fix only one issue at a time with self explanatory comment.
I know as I have a reproducer and have tested that this PR indeed fixes my reproducer. Which sithlord unrelated PR likely does not.

Though I agree that sithlord deskflow PR to avoid triggering the input emulation on borders that have no side panels and the removal of the FakeEIS call are worthy but are totally unrelated to this PR and to the bug report it closes.

nbolton:
Ah, @prahal is right; this PR ports a fix from upstream originally by @mrnicegyu11:
https://github.com/deskflow/deskflow/pull/8503 (more specifically: https://github.com/deskflow/deskflow/pull/8503/commits/1b8c24e7ed2c0c6d29ede5185412b7c55d1d4986)

Upstream sponsored the fix as a $3000 bounty.

That PR was indeed the fix for the issue referenced in OP:
- https://github.com/deskflow/deskflow/issues/8005

@sithlord48 FYI ^

Edit: @prahal You may also want to port this fix from upstream if applicable:
- https://github.com/deskflow/deskflow/pull/8503/commits/f403945bc7961440d759b50a3a4ac81b36ccaf3e

sithlord48:
> Ah, @prahal is right; this PR ports a fix from upstream originally by @mrnicegyu11: [deskflow/deskflow#8503](https://github.com/deskflow/deskflow/pull/8503) (more specifically: [deskflow/deskflow@1b8c24e](https://github.com/deskflow/deskflow/commit/1b8c24e7ed2c0c6d29ede5185412b7c55d1d4986))

This will only fix #2287 in cases where the portal backend has crashed or been otherwise killed. The other case is when you hit an edge that is marked with an EIS border but has no screen on the other side. To fully fix this (as we did in upstream) you should add https://github.com/deskflow/deskflow/pull/8749/commits/a4e2a1831ecb4634d3e6c865aac648459edd7cf6 to this PR.

nbolton:
> when you hit an edge

Argh, I wasn't looking at #2287 -- yeah, that issue is a different bug.

Should OP change "Closes #2287" to something else?

sithlord48:
I guess since this would not fully close that issue as is.

prahal:
> > Ah, @prahal is right; this PR ports a fix from upstream originally by @mrnicegyu11: [deskflow/deskflow#8503](https://github.com/deskflow/deskflow/pull/8503) (more specifically: [deskflow/deskflow@1b8c24e](https://github.com/deskflow/deskflow/commit/1b8c24e7ed2c0c6d29ede5185412b7c55d1d4986))
> 
> This will only fix #2287 in cases where the portal backend has crashed or been otherwise killed. The other case is when you hit an edge that is marked with an EIS border but has no screen on the other side. To fully fix this (as we did in upstream) you should add [deskflow/deskflow@a4e2a18](https://github.com/deskflow/deskflow/commit/a4e2a1831ecb4634d3e6c865aac648459edd7cf6) to this PR.


@sithlord46 I disagree. A PR ought to fix one bug and be commented. Not all bug related to mouse. Not only is it not a good idea to merge multiple fixes in one PR, it is wrong. It breaks bisecting regressions.

Edit: Well in fact I mean it is wrong to merge multiple  fixes in one commit not in one PR. You can have as many bug fixes in one PR as long as they are split in different commits.

But in my opinion the edge issue has nothing to do with the issue at stack in the issue I opened which is related to this PR, that is #2287.
We should not have PR that apply all possible fixes for EIS in one PR. Even if it is not wrong per se if they are split in different commits. github history will get tangled if we close many issue in one PR, even if not wrong per se it is confusing and has no added value.

prahal:
> I guess since this would not fully close that issue as is.

But it fully closes #2287 as is... what issue does this PR not fully close? Mind I mentioned the deskflow issue as a reference to get more details about how to reproduce the bug and how it got where it is. I never intended to close the deskflow issue with this PR. This PR is about #2287 only.
I am only keeping the deskflow backtrack to keep authorship where due. Not because this PR is about the deskflow issue I mentioned.

sithlord48:
> @sithlord46 I disagree. A PR ought to fix one bug and be commented. Not all bug related to mouse.

You can disagree all you like but I am telling you this as a matter of fact. It does not fix the bug you reported. It fixes one of two possible ways to trigger it.  Either add the fix for the second case or don't close the bug when landing this since there is a second known way to trigger it. 

> Not only is it not a good idea to merge multiple fixes in one PR, it is wrong. 

Its possible one for a bit of incorrect code to manifest as several seemingly unrelated bugs. That is exactly the case with the deskflow commit i linked. Let me explain . When EI sets the edges they are expected to go to another screen.. if they dont here is what can happen  
1. The cursor hitting an edge can be banished to no where. This is the other cause of the bug your trying to fix.
2. The cursor will start to blink on the edge. This is a second bug , Why does it happen EI tells the compositor the cursor is going to another screen the cursor get hidden when libEI has no where to move the cursor it reappears.
3. The cursor blocked by EI from hitting the true edge as a result any gestures trigged by such an action become impossible. That is a third bug. 

> It breaks bisecting regressions.

 How bisect goes  commit by commit in range. Add the other commit as a commit to this PR and you can bisect between them .

prahal:
> > @sithlord46 I disagree. A PR ought to fix one bug and be commented. Not all bug related to mouse.
> 
> You can disagree all you like but I am telling you this as a matter of fact. It does not fix the bug you reported. It fixes one of two possible ways to trigger it.  Either add the fix for the second case or don't close the bug when landing this since there is a second known way to trigger it. 
> 
> > Not only is it not a good idea to merge multiple fixes in one PR, it is wrong. 
> 
> Its possible one for a bit of incorrect code to manifest as several seemingly unrelated bugs. That is exactly the case with the deskflow commit i linked. Let me explain . When EI sets the edges they are expected to go to another screen.. if they dont here is what can happen  
> 1. The cursor hitting an edge can be banished to no where. This is the other cause of the bug your trying to fix.
> 2. The cursor will start to blink on the edge. This is a second bug , Why does it happen EI tells the compositor the cursor is going to another screen the cursor get hidden when libEI has no where to move the cursor it reappears.
> 3. The cursor blocked by EI from hitting the true edge as a result any gestures trigged by such an action become impossible. That is a third bug. 
> 
> > It breaks bisecting regressions.
> 
>  How bisect goes  commit by commit in range. Add the other commit as a commit to this PR and you can bisect between them . 

You misunderstood. I was talking about the fact my push request ought not to include both the already included fix, the fix for the sides without a screen on the edge, and the fix that removes FakeEIS. Your commit does not include the commit I backported here, and the commit you got merged. What I mean is you are asking for this push request to include all and every code fix related to EIS, which is bad. You did not do it. You are asking me to do it or this push request should in your opinion not be merged. I disagree.

About push request, common habits, I spent some time grabbing guidance guides written by others. You are free to agree or disagree, but these are rules that were defined even before git was created, and they help multiple people work on the same repository. Your call. Here is one:
```
Thanks to some guidance from my peers, they suggested that while my code was valid, I wasn’t doing myself any favors by lumping my work into a single massive pull request. Instead, I should break a feature down into several easy-to-understand, easy-to-roll-out pull requests. Adopting this recommendation was a notable turning point in my journey as a developer. Ever since, my pull requests have few comments and change requests, get approved sooner, and ship to customers faster. As a result, my development process feels more collaborative and rewarding rather than confrontational and frustrating.
``` from https://solace.com/blog/pull-request-best-practices/

Well, this rule is for push requests, which by definition is at times wider than the rules for commits.
I want to work with you even if you refuse to abide by common habits. But I wanted to share the advice I was given and which in my opinion is good.

**Comment by prahal** [comment]
> > @sithlord46 I disagree. A PR ought to fix one bug and be commented. Not all bug related to mouse.
> 
> You can disagree all you like but I am telling you this as a matter of fact. It does not fix the bug you reported. It fixes one of two possible ways to trigger it.  Either add the fix for the second case or don't close the bug when landing this since there is a second known way to trigger it. 

I doubt I cannot know that the bug I reported is fixed when I am the one reporting the bug. I mean how can you know what was exactly my issue ? And that it has nothing to do with edges that had no screen attached?
Also I believe the FakeEIS removal is unrelated. But I cannot test that since with the code in this push request I cannot reproduce input freeze in any way. Probably I never trigger the reproducer for the bug you have fixed.

**Comment by prahal** [comment]
> > Not only is it not a good idea to merge multiple fixes in one PR, it is wrong. 
> 
> Its possible one for a bit of incorrect code to manifest as several seemingly unrelated bugs. That is exactly the case with the deskflow commit i linked. Let me explain . When EI sets the edges they are expected to go to another screen.. if they dont here is what can happen  
> 1. The cursor hitting an edge can be banished to no where. This is the other cause of the bug your trying to fix.
> 2. The cursor will start to blink on the edge. This is a second bug , Why does it happen EI tells the compositor the cursor is going to another screen the cursor get hidden when libEI has no where to move the cursor it reappears.
> 3. The cursor blocked by EI from hitting the true edge as a result any gestures trigged by such an action become impossible. That is a third bug. 
> 
> > It breaks bisecting regressions.
> 
>  How bisect goes  commit by commit in range. Add the other commit as a commit to this PR and you can bisect between them . 

Because many issue with mangling multiple functional code change. If you think you know better than any other person that was trained this way, you can proceed how you always had. Nobody can force you to code otherwise you like on a project you are a free contributor about, well except projects owners. But I can assure you this is how everybody is trained.
I have bisected quite a lot and found these rules efficient. Could be yours are good too. But these ones have been proven on the field.

 I don't mean you break bisect from going back and forth in the git history. I am saying the point of bisecting is to test each commit for the regression. If you have too much code or many fixes in a single commit, bisect will still "work" but the point of bisecting will not work anymore. That is you might have multiple regression in a single commit, thus bisecting has no point anymore.
Because one regression might hide the other.

You can even end up a single commit with one regression then another regression to make the first regression "vanish" but that is work but with a broken code... until someone fix the code years later with a correct fix but then the first regression reappears. This is impossible to bisect if you mix multiple fixes in one commit.
As I told, I cannot force you to do as is taught to every developer. And you might never be negatively impacted in your whole life by the way you commit. These are just conventions aimed at easing collaborative work.

**Comment by prahal** [comment]
> The cursor hitting an edge can be banished to no where. This is the other cause of the bug your trying to fix.

```
It can happen that the libeis implementation (read: the wayland compositor) sends a start_emulating event to a device already emulating. This is a protocol violation of libei/libeis. The client<-->server connection is severed (which is correct behavior), but deskflow never releases the captured xdg-desktop-portal-input-capture. Thus, the cursor remains perpetually stuck. The fix is to release the portal-input-capture when deskflow handles the EI_EVENT_DISCONNECT event.
```
 https://github.com/deskflow/deskflow/pull/8503

```
>    sends a start_emulating event to a device already emulating.

fwiw this really shouldn't be possible using the libeis API, can you file a bug in libei please and I'll have a look at it.

This issue may also hint at a bug in the compositor because if the ei connection is terminated it should send the Closed signal and this issue should've never shown up in deskflow. As it is from a review of the patches (I've paged most of the details out by now, sorry) this approach looks correct though. Do make sure though all the signals pertaining to the current session are disconnected too (probably done in the destructor anyway, but worth double-checking) - this way we won't get a delayed Closed event on an already destroyed session.
```
https://github.com/deskflow/deskflow/pull/8503#issuecomment-2822773891

How has this anything to do with edge without screens?
Or FakeEISFd removal?

It can even be fixed on the compositor side only. But we should close our connection to be good citizen that follows API thus this PR.

prahal:
> > when you hit an edge
> 
> Argh, I wasn't looking at #2287 -- yeah, that issue is a different bug.
> 
> Should OP change "Closes #2287" to something else?

@nbolton I don't see how that issue is a different bug. Note that I am both the #2287 issue reporter and this PR reporter.
When an issue tells the application crashes when you move the mouse, you are not only allowed to close the issue when all bug related to mouse are fixed in this PR... you can close it when the reporter specific mouse bug is fixed.

In my bug report, I mention that one have to mouse the mouse between both the server and the client screen. To me, this cannot relate to moving the mouse to side without clients. Or at least that was the intent of the sentence I wrote when I reported the issue.

You can easily reproduce my issue with current input-leap master and check that it is fixed when you apply this PR.

p12tic:
@prahal Thanks a lot and sorry for half a year delay..

Issue #2348, opened 2026-01-11
Misc. issues trying to build and debug input-leap for drag and drop breakage discussed in #258 / #855 on Windows


deliciouslytyped:
Related to: #258 / #855 / debauchee/barrier/#855

I apologize in advance for the disorganization in this issue, this is partly just notes for myself regarding what I did.

Note https://github.com/deskflow/deskflow/discussions/8082 suggests an alternative tool for local file transfers.
> Edit: Drag and drop was removed and is no longer supported. Consider using LocalSend.
( I've currently started testing KDE Connect, which also works on windows. - but it doesn't look at all as seamless as input-leap drag should be. https://userbase.kde.org/Tips/Transfer_files_between_phone_and_PC_using_Wi-Fi `On any platform other than iOS, your file will be transferred in the background to the Downloads folder of the receiving device. You can use your platform's file manager to find it. You can also check your device’s notifications for more information about the progress. `)

I'm using Windows 10 19045.6456 for input-leap server.

#### build stuff
I really didn't want to deal with installing visual studio and building from source, so some issues I ran into are related to that, but here we are.

- the input-leap windows debug release at least as of 3.0.3 does not distribute pdbs
- windows debug builds need visual studio installed to provide debug dlls such as msvcp140d.dll mscvp140_1d.dll, ucrtbased.dll and/or others (somewhat explained at https://learn.microsoft.com/en-us/visualstudio/releases/2019/redistribution). I have not found any way to acquire these without installing most of the visual studio c++ components, including the windows sdk for ucrtbased.dll.
- qt company now seems to require logging in to get binary builds of qt, which I don't want to do. vcpkg apparently provides a way to build from source in an automated fashion, but with it's own downsides. (takes a lot of time/space). this also needs the visual studio c++ cmake component installed to not error. I'm currently waiting for this to finish to see if I can use it to build input-leap.

I'm following https://github.com/input-leap/input-leap/wiki/Building-on-Windows to try building from source.

Disclaimer: I have no idea how to *correctly* use vcpkg.
Apparently, this CMAKE define must be set (in the GUI, *before* the first generation/configure): `CMAKE_TOOLCHAIN_FILE=<<someinstalldir>>\vcpkg\scripts\buildsystems\vcpkg.cmake`
(Docs apparently at https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration . I didn't read them.)
[...]

Issue #855, opened 2021-11-04
Tracker issue: Drag and Drop.


ghost:
> This issue has been migrated from old Barrier Github repository https://github.com/debauchee/barrier/issues/855
> 
> Issue created on: 2020-08-28 by @shymega
> Issue last updated on: 2021-10-14

Due to the numerous issues created, without any sign of checking for duplicate issues, I have taken this opportunity to _close_ those issues about drag and drop not working in favour of a **centralised** tracking issue.

Please only comment on this issue with constructive and positive messages that will facilitate the development and maintenance of drag and drop support. Unconstructive messages will be removed if they do not fit the discussion. Reacting to messages is fine, however, as it indicates the general feeling towards construction and maintenance of drag and drop.

Thank you :smile:

---

> Commented on: 2020-08-28 by @shymega

On the note of reactions, could those with this issue, please react to this comment with :+1: (`:+1`). Thanks. This will allow us to gauge the spread of the issue.

---

> Commented on: 2020-09-14 by @chrishultberg

Using Ubuntu 18.04.3 LTS as the server and Windows 10 Pro as the client, I can not drag and drop files.
This also means that I can't copy/paste the files...

Doesn't seem like there's been an update to this in a few weeks so I wanted to double-check.

---

> Commented on: 2020-09-15 by @shymega

You don't need to double-check. If the issue hasn't been updated, please assume nothing has changed. Which it hasn't. I only say this so the tracker issue doesn't get "noisy" with comments asking if there's been any progress etc....

---

> Commented on: 2020-11-27 by @arthur-tacca

Personally, I would prefer a file transfer that was very clunky (even just pasting the source and destination filenames into text boxes) but *worked*, rather than something with a smooth interface (like drag and drop) but is unreliable. This is especially the case with Linux - I don't believe drag and drop is even intended to work on Linux but Linux-Windows is my primary use case, so anything at all would be a great help with that.

(Apologies for commenting on this high-activity thread but I think this is genuinely new content. Those that agree/disagree with my preference had add their reactions accordingly to this comment.)

---

> Commented on: 2021-05-22 by @shymega

All - this is a _tracking_ issue for the development of this feature. Please only comment constructively. I'm going to delete those comments that merely post logs etc of the current behaviour. I recognise my decision may be controversial, but we really need to keep the issue constructive. Thank you.

---

> Commented on: 2021-05-23 by @shymega

[...]

PR #2349, opened 2026-01-11
Fix drag and drop with windows GUI+service usage mode

deliciouslytyped:
Related: #2348 / #855 / debauchee/barrier/#855

This is a PR accompanying #2348 .
I did the debugging myself. I'm not familiar with the Windows API, so I did not write the code myself, or review it in detail.

**This commit contains LLM generated code, is likely to contain bugs, and should not be merged as is.** It serves as a proof of concept, but probably needs a rewrite.

See the extensive commit messages for information about what's going on.

## Contributor Checklist:


deliciouslytyped:
A second commit has been added fixing the sender side. See the extensive commit message for additional information.

deliciouslytyped:
Apparently something still needs to be tuned here, because the drag capturing is not completely reliable.

It would probably be better not to spawn a separate process as well, but this was the easiest way forward that I could see.

deliciouslytyped:
Apparently something still needs to be tuned here, because the drag capturing is not completely reliable.

It would probably be better not to spawn a separate process as well, but this was the easiest way forward that I could see.

Also, switching screens seems to interrupt transfers, and apparently it's not happy with large transfers. (I guess it uses the clipboard?)
It would be good to know if the current expected behavior is documented anywhere.

deliciouslytyped:
Also, the uiaccess workaround for UIPI restrictions which is related to accessibility tools or something, would be an alternative solution to some things, but has some additional restrictions, which would cause some annoyance:
https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/user-account-control-allow-uiaccess-applications-to-prompt-for-elevation-without-using-the-secure-desktop
````
If an application presents a UIAccess attribute when it requests privileges, the application is stating a requirement to bypass UIPI restrictions for sending messages across privilege levels. Devices implement the following policy checks before starting an application with UIAccess privilege.

1.    The application must have a digital signature that can be verified by using a digital certificate that is associated with the Trusted Root Certification Authorities store on the local computer.

 2.   The application must be installed in a local folder that is writeable only by administrators, such as the Program Files directory. The allowed directories for UI automation applications are:
        %ProgramFiles% and its subdirectories.
        %WinDir% and its subdirectories, except a few subdirectories that are excluded because standard users have write access.

```

PR #2360, opened 2026-02-21
fix: Raise main window when showing it

jcs:
(On macOS at least) when selecting 'Show' from the tray icon's menu, the main window shows but is always behind the current application which is confusing.

## Contributor Checklist:

* [X] This change affects end users and I have created a file in the `doc/newsfragments` directory (and made sure to read the `README.md` in that directory)


sithlord48:
#2199  , This has been fixed [upstream](https://deskflow.org)

PR #2372, opened 2026-04-14
fix(wayland): add missing HAVE_LIBPORTAL_INPUTCAPTURE guard in EiScreen.cpp

srlinuxme:
FIX EiScreen.cpp

Commit b7d1242c introduced portal_input_capture_ usage in the EI_EVENT_DISCONNECT handler without the required preprocessor guard. On Ubuntu 24.04/24.10, libportal 0.7.1 does not provide xdp_input_capture_session_connect_to_eis, so HAVE_LIBPORTAL_INPUTCAPTURE is not defined, causing a build failure when INPUTLEAP_BUILD_LIBEI=1.

## Contributor Checklist:


sithlord48:
#2199

PR #2373, opened 2026-04-14
feat(gui): use monochrome white tray icons on Linux/GNOME

srlinuxme:
Ubuntu GNOME's system tray expects monochrome (template) icons. This change adds:

- White PNG variants of the existing mask icons as embedded fallback
- Freedesktop symbolic SVG icons installed to share/icons/hicolor/symbolic/apps/ (connected, disconnected, transfering)
- On Linux, QIcon::fromTheme uses the "-symbolic" naming convention so GNOME Shell renders them as monochrome automatically
- Fallback to the embedded white PNGs when the theme is not available
- /debian/ and /builddir/ added to .gitignore (generated by CI)

## Contributor Checklist:


sithlord48:
#2199

PR #2376, opened 2026-04-28
Fix/malfunc wayland

srlinuxme:
## Contributor Checklist:


sithlord48:
#2199 , Maybe better off trying [upsteam](https://github.com/deskflow/deskflow) and seeing if these issues are still around. If so we do accept PRs

PR #2378, opened 2026-05-07
feat: add low-latency input mode and fix macOS mouse bugs

liuli0212:
## Summary

- Add `INPUTLEAP_LOW_LATENCY_INPUT` environment variable to enable low-latency input mode, which disables mouse motion compression and delivers key/mouse events immediately
- Fix macOS mouse coordinate typo (`mx == m_yCenter` → `my == m_yCenter`)
- Filter zero integer relative mouse moves on macOS to avoid unnecessary events
- Reduce XWindows poll timeout from 25ms to 1ms
- Re-enable `XAutoRepeatOn`/`XAutoRepeatOff` on XWindows
- Default `cryptoEnabled` to `false`

## Background

Tested in a Mac (WiFi) + Linux (wired) environment. The mouse cursor movement is very laggy/choppy, especially noticeable during fast mouse movements. This PR addresses the latency issues by:

1. **Mouse motion compression** was batching events and adding noticeable delay — the new low-latency mode skips compression entirely
2. **Key event delivery** was not immediate — now uses `Event::kDeliverImmediately` flag when enabled
3. **XWindows poll timeout** of 25ms was too high — reduced to 1ms
4. **macOS mouse coordinate bug** caused incorrect boundary checks, leading to missed or extra motion events
5. **Zero relative moves** were being sent unnecessarily, adding overhead

## Usage

Enable low-latency mode by setting any of these environment variables before starting InputLeap:

```bash
export INPUTLEAP_LOW_LATENCY_INPUT=1
# or
export INPUTLEAP_LOW_LATENCY_MOUSE=1
```

To explicitly disable (default behavior without env var):
```bash
export INPUTLEAP_DISABLE_LOW_LATENCY_INPUT=1
```

## Note

This PR is intended as a **reference implementation** for guys with similiar problems as mine. The changes are opt-in via environment variable and do not affect default behavior (except the reduced XWindows poll timeout and the macOS bug fixes).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


sithlord48:
#2199

